PR #3, opened 2020-08-18
Initial python iothub device client test app

BertKleewein:


docs/schema.md
```diff
@@ -16,7 +16,7 @@
 Desired:
 ```json
 {
-  "longhaulTelemetrySendInterval": 10
+  "longhaulTelemetrySendIntervalInSeconds": 11
```
drwill-ms, line 19: Could probably drop the `longhaul` prefix as everything in this repo is long-haul, right?

python/device/d2c.py
```diff
@@ -0,0 +1,336 @@
...
+import uuid
+import json
+import datetime
+from measurement import ThreadSafeCounter, ThreadSafeList, MeasureLatency
+from concurrent.futures import ThreadPoolExecutor
+
+from azure.iot.device import IoTHubDeviceClient, Message
+
+logging.basicConfig(level=logging.ERROR)
+# logging.getLogger("paho").setLevel(level=logging.DEBUG)
```
drwill-ms, line 18: any logging for application insights?

python/device/d2c.py
```diff
@@ -0,0 +1,336 @@
...
+
+            t.average_d2c_roundtrip_latency_to_gateway_in_seconds = (
+                self.metrics.d2c.latency.extract_average()
+            )
+            t.d2c_in_flight_count = self.metrics.d2c.inflight.get_count()
+            t.d2c_success_count = self.metrics.d2c.succeeded.extract_count()
+            t.d2c_failure_count = self.metrics.d2c.failed.extract_count()
+
+            msg = Message(json.dumps(t.to_dict()))
+            msg.content_type = "appliucation/json"
```
drwill-ms, line 213: misspelled application

python/device/measurement.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class MeasureLatency(contextlib.AbstractContextManager):
+    def __init__(self, tracker=None):
+        self.start_time = None
+        self.end_time = None
+        self.tracker = tracker
+
+    def __enter__(self):
+        self.start_time = datetime.datetime.now()
```
drwill-ms, line 79: utc?

BertKleewein: Good point  We should always use UTC when pushing data up.  This class is about measuring latency.  I may even change this to use epoch time since we only care about deltas in seconds.  

---
In reply to: [474140237](https://github.com/Azure/iot-sdk-longhaul/pull/3#discussion_r474140237) [](ancestors = 474140237)

PR #4, opened 2020-08-20
Early .net device impl

drwill-ms:
I've added the beginnings of an app framework to manage the connection, have independent classes that handle telemetry for specific kinds of app logic, and an exit condition. Includes some basic logging to appinsights.

### DIFF
--- a/net/device/IIotHub.cs
+++ b/net/device/IIotHub.cs
@@ -1,0 +1,26 @@
+﻿using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Iot.Thief.Device
+{
+    /// <summary>
+    /// An interface for device code to interact with hub
+    /// </summary>
+    internal interface IIotHub
+    {
+        /// <summary>
+        /// Sends the specified telemetry object as a message.
+        /// </summary>
+        /// <param name="telemetryObject">An object to be converted to an application/json payload to send.</param>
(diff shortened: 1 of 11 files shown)


net/device/IIotHub.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Iot.Thief.Device
+{
+    /// <summary>
+    /// An interface for device code to interract with hub
+    /// </summary>
+    internal interface IIotHub
```
drwill-ms, line 9: to be added:
- write properties
- hook to respond to registered C2D messages
- hook to respond to registered commands

net/device/IotHub.cs
```diff
@@ -0,0 +1,157 @@
...
+                foreach (var prop in extraProperties)
+                {
+                    iotMessage.Properties.TryAdd(prop.Key, prop.Value);
+                }
+            }
+            await _deviceClient.SendEventAsync(iotMessage).ConfigureAwait(false);
+            _logger.Trace($"Sent message [{message}]");
+        }
+
+        private void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
```
drwill-ms, line 96: Based off Abhipsa's implementation in the reconnect sample. I haven't yet added the exception handling logic, nor have I made use of the _isConnected variable (pretty sure I want to prevent telemetry from being sent when retrying, but not sure what that looks like yet).

PR #7, opened 2020-09-12
More impl, batch telemetry

drwill-ms:


net/device/ExceptionHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        private static readonly HashSet<Type> s_networkExceptions = new HashSet<Type>
+        {
+            typeof(IOException),
+            typeof(SocketException),
+            typeof(ClosedChannelException),
+            typeof(TimeoutException),
+            typeof(OperationCanceledException),
+            typeof(HttpRequestException),
+            typeof(WebException),
+            typeof(WebSocketException),
```
BertKleewein, line 30: This list is interesting to me.  Elena wants to track how often different failures occur.  I wonder what granularity we want to support -- breaking down network exceptions into these 8 types could certainly be useful.  The tricky part is coming up with a way to be (relatively) consistent across languages with how we map failures into these groups.

net/device/IotHub.cs
```diff
@@ -45,7 +56,8 @@ public async Task InitializeAsync()
 
             try
             {
-                if (Cleanup())
+                if (_deviceClient == null
+                    || Cleanup())
```
abhipsaMisra, line 60: q -> the term "cleanup" makes me think that the application is carrying out the "cleanup" stuff before shutting down; the fact that we are initializing the client instance inside this block was a bit confusing for me. Can we rename this to something like "WasClientCleanedUp"?

net/device/IotHub.cs
```diff
@@ -75,10 +87,97 @@ public bool Cleanup(bool force = false)
...
+                // If not connected, skip the work below this round
+                if (!_isConnected)
+                {
+                    _logger.Trace($"Waiting for connection before batching telemetry", TraceSeverity.Warning);
+                    continue;
+                }
+
+                // Make a batch of messages to send, unless we're retrying a previous batch
+                if (!_pendingMessages.Any()
+                    && _messagesToSend.Any())
```
abhipsaMisra, line 120: In case of long periods of disconnect, `_messagesToSend` will keep on getting bigger, without any message being actually sent out. Should we be worrying about storage space constraints?

net/device/IotHub.cs
```diff
@@ -75,10 +87,97 @@ public bool Cleanup(bool force = false)
...
+                    {
+                        _logger.Trace($"Caught transient exception; will retry: {ex}", TraceSeverity.Warning);
+                    }
+                    catch (Exception ex) when (ExceptionHelper.IsNetworkExceptionChain(ex))
+                    {
+                        _logger.Trace($"A network-related exception was caught; will retry: {ex}", TraceSeverity.Warning);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Trace($"Unknown error batching telemetry: {ex}", TraceSeverity.Critical);
```
abhipsaMisra, line 169: not related to batching logic - since we consume all exception types, some kind of a counter here will keep track of how many times we got a network exception, etc. 
Also, we might want to add an exit strategy -> if we get a critical exception more than 5 times in a row, we should exit/ log exception and then dispose the existing client and recreate?

abhipsaMisra (approved):
Batching logic looks good

timtay-microsoft:
Are we switching to batching messages because we expect it will perform better for longhaul tests? Or is this just a more common user scenario that we want to optimize on?

net/device/MessageHelpers.cs
```diff
@@ -0,0 +1,37 @@
...
+    internal static class MessageHelpers
+    {
+        /// <summary>
+        /// Reported max of entire payload is 256k, but we'll hold back 1 k as fudge factor for things we can't calculate.
+        /// </summary>
+        public const int MaxMessagePayloadBytes = (256 * 1024) - 1024;
+
+        private const int Iso8601Length = 24;
+
+        public static int GetMessagePayloadSize(Message message)
```
BertKleewein, line 16: I know this is muddy because of iothub mixing byte count and character count, but here you use the suffix "Size" and above you use the suffix "Bytes" in MaxMessagePayloadBytes.  This leads to 2 concerns:
1. Is the limit really 255K bytes or is is really 255k characters? 
2. If "character" is the unit of measurement that we're using, that should be explicit, maybe in the function name.
I know that C# is safer than C, but mixing "count of bytes" and "count of characters" is a common mistake and a great way to introduce buffer overruns.

net/device/ExceptionHelper.cs
drwill-ms: I copied it from Abhipsa's new sample https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/DeviceReconnectionSample

net/device/IotHub.cs
drwill-ms: I don't see it the same way, but you did make me take a new look at it. I'll make some changes.

net/device/IotHub.cs
drwill-ms: I'm not worrying about memory constraints until I find we have to. I concur it is a potential problem though.

net/device/IotHub.cs
drwill-ms: Part of the point of long-haul testing is to find these scenarios. In this case, I don't want to proactively code for a condition we haven't found.

net/device/MessageHelpers.cs
drwill-ms: The mix of characters and bytes isn't my mistake, but I am staying true to the mistake that was made.

net/device/ExceptionHelper.cs
abhipsaMisra: Just FYI - these are the exception types that the .NE sdk treats as transient network exceptions: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/ErrorDelegatingHandler.cs#L29

net/device/IotHub.cs
drwill-ms: Updated. What do you think?

abhipsaMisra: yup, I like it

PR #13, opened 2020-10-08
Add Azure App Analytics, refactor again.

BertKleewein:


BertKleewein:
This adds code to push logs and events to Azure Monitor (but not metrics yet).  The code was also refactored (again) , because I wanted to build code that didn't rely on monumental base classes.   I did this because I want a sample app which can easily be copy/pasted and base classes make that difficult to accomplish.

python/containers/fetch-device-secrets.sh
```diff
@@ -10,5 +10,6 @@ echo Fetching secrets
 get-secret THIEF_DEVICE_PROVISIONING_HOST THIEF-DEVICE-PROVISIONING-HOST
 get-secret THIEF_DEVICE_ID_SCOPE THIEF-DEVICE-ID-SCOPE
 get-secret THIEF_DEVICE_GROUP_SYMMETRIC_KEY THIEF-DEVICE-GROUP-SYMMETRIC-KEY
+get-secret THIEF_AI_CONNECTION_STRING THIEF-AI-CONNECTION-STRING
```
YoDaMa, line 13: AI lol

python/service/service.py
```diff
@@ -174,6 +206,7 @@ def c2d_sender_thread(self):
                             )
                             with self.registry_manager_lock:
                                 del self.registry_manager
+                                time.sleep(1)
```
YoDaMa, line 209: is there a reason for this random delay

YoDaMa (approved):

python/containers/fetch-device-secrets.sh
BertKleewein: AI - Application Insights.  I think maybe I need to rename this variable 

---
In reply to: [501948558](https://github.com/Azure/iot-sdk-longhaul/pull/13#discussion_r501948558) [](ancestors = 501948558)

python/service/service.py
BertKleewein: This is in a error-recovery block of code, and the delay seemed prudent to let the old registry_manager instance clean itself up.  I was hitting this pretty consistently at some point and I'm not anymore. Either this recovery code does the job or some other bug was fixed.

---
In reply to: [501950818](https://github.com/Azure/iot-sdk-longhaul/pull/13#discussion_r501950818) [](ancestors = 501950818)

PR #14, opened 2020-10-13
align with .net impl, support many:one service-to-device app ratios

BertKleewein:
* Aligned telemetry names with .net
* refactored system_health_telemetry.py to match .net
* removed heartbeat.  we decided that this was too extreme and a different mechanism to deal with missing partners will be added later.
* changed code so we can have a single service app that works with multiple device apps.
* added "bootstrap" mechanism that a device can use to pair itself with a service app.
* added environmental variables for run_id and a way for a device app to pair with a specific service app.

### DIFF
--- a/python/device/device.py
+++ b/python/device/device.py
@@ -23,17 +23,30 @@ event_logger = get_event_logger("device")
 
 logging.basicConfig(level=logging.INFO)
 logging.getLogger("paho").setLevel(level=logging.DEBUG)
-logging.getLogger("thief").setLevel(level=logging.INFO)
+logging.getLogger("thief").setLevel(level=logging.DEBUG)
 
 
+# use os.environ[] for required environment variables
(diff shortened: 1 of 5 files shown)


python/device/device.py
```diff
@@ -139,11 +141,80 @@ def update_initial_reported_properties(self):
...
+                    reported_properties["thief"]["device"]["serviceAppRunId"] = service_app_run_id
+                    self.client.patch_twin_reported_properties(reported_properties)
+                    return
+                else:
+                    logger.debug("patch received, but serviceAppRunId not set: {}".format(patch))
+
+            logger.debug("Still no assigned service instance. Sending new bootstrap message")
+            self.client.send_message(bootstrap_message)
+
+        raise Exception("timeout waitig for service to set serviceAppRunId")
```
drwill-ms, line 216: waiting

drwill-ms (approved):

PR #15, opened 2020-10-15
add azure monitor metrics to python and change schema to be closer to .net

BertKleewein:


python/device/device.py
```diff
@@ -108,39 +115,100 @@ def __init__(self):
...
+        self.reporter.add_integer_measurement(
+            "process_working_set_private",
+            "processWorkingSetPrivate",
+            "Amount of non-shared physical memory used by the process",
+            "bytes",
+        )
+        self.reporter.add_integer_measurement(
+            "total_messages_sent",
+            "totalMessagesSent",
+            "count of messages sent and ack'd by the transport",
```
drwill-ms, line 156: This and several below don't capitalize the first word

python/device/device.py
```diff
@@ -108,39 +115,100 @@ def __init__(self):
...
+        self.reporter.add_integer_measurement(
+            "messages_in_flight",
+            "messagesInFlight",
+            "count of messages sent to iothub but not ack'd by the transport",
+            "message(s)",
+        )
+        self.reporter.add_integer_measurement(
+            "messages_unacknowledged",
+            "messagesUnacknowledged",
+            "count of messages sent to iothub and acked by the transport, but receipt not verified via service sdk",
```
drwill-ms, line 180: I'm confused, there are 2 levels of acknowledgment, transport and service?

FWIW, we don't seem to have this intermediate stage in .NET or Java, at least how the API is designed.

python/device/device.py
```diff
@@ -108,39 +115,100 @@ def __init__(self):
...
-                "requestsSent": self.metrics.pingback_requests_sent.get_count(),
-                "responsesReceived": self.metrics.pingback_responses_received.get_count(),
-            },
-            "d2c": {
-                "queued": self.telemetry_queue.qsize(),
-                "inFlight": self.metrics.d2c_in_flight.get_count(),
-                "sent": self.metrics.d2c_sent.get_count(),
-                "acknowledged": self.metrics.d2c_acknowledged.get_count(),
-            },
+            "totalMessagesSent": sent,
```
drwill-ms, line 198: I'm sending these values as telemetry only, not as props.

drwill-ms: Err, I mean to AppInsights. I could see sending them as telemetry though. They don't seem like values worth of twin props to me. What do you think?

python/device/device.py
BertKleewein: In my next PR, I've renamed these to acked and ReceeivedByService.

---
In reply to: [505741511](https://github.com/Azure/iot-sdk-longhaul/pull/15#discussion_r505741511) [](ancestors = 505741511)

python/device/device.py
BertKleewein: I'm putting them into twin props because it gives me an easy way to check status without doing an app insights query.  We can remove these later.

---
In reply to: [505743261](https://github.com/Azure/iot-sdk-longhaul/pull/15#discussion_r505743261) [](ancestors = 505743261)

PR #17, opened 2020-10-27
standardize config, add error checking

BertKleewein:
* Move all of the config parameters into a single structure which will be easier to move into writable properties in the future.  
* Rename metrics for consistent naming 
* Add error checking configuration and code for send_message failure conditions.


python/device/device.py
```diff
@@ -78,16 +78,47 @@ class DeviceRunConfig(object):
     Currently hardcoded. Later, this will come from desired properties.
     """
 
+    # All durations are in seconds
     def __init__(self):
-        self.max_run_duration = 90
+        # how long should teh test run before finishing.  0 = forever
```
drwill-ms, line 83: misspelled 'the'

python/device/device.py
```diff
@@ -78,16 +78,47 @@ class DeviceRunConfig(object):
...
+        self.pairing_request_timeout_interval = 900
+
+        # How many seconds to wait before sending a new pairingRequest message
+        self.pairing_request_send_interval = 30
+
+        # How many times to call send_message per second
+        self.send_message_operations_per_second = 1
+
+        # How many threads do we spin up for overlapped send_message calls
+        self.send_message_thread_count = 10
```
drwill-ms, line 102: Can you elaborate? Does this mean, if we're disconnected, we'll have, at max, 10 events waiting to send? That means if we disconnect longer than 10 seconds, we start to lose events that otherwise would have been sent?

python/device/device.py
```diff
@@ -78,16 +78,47 @@ class DeviceRunConfig(object):
...
+
+        # How long do we wait for notification that the service app received a
+        # message before we consider it a failure?
+        self.send_message_arrival_failure_interval = 3600
+
+        # How many messages do let not arrive at the service before we fail the test
```
drwill-ms, line 108: "do let"?

python/device/device.py
```diff
@@ -380,14 +429,19 @@ def on_pingback_received(pingback_id):
                 self.metrics.d2c_received_by_service_app.increment()
 
             with self.pingback_list_lock:
-                self.pingback_wait_list[pingback_id] = (on_pingback_received, (pingback_id,))
+                self.pingback_wait_list[pingback_id] = MessageWaitingToArrive(
+                    callback=on_pingback_received,
+                    pingback_id=pingback_id,
+                    message=msg,
+                    send_time=time.time(),
```
drwill-ms, line 436: is this utc time?

python/device/device.py
```diff
@@ -457,18 +511,90 @@ def handle_pingback_thread(self, worker_thread_info):
...
+                arrival.callback(arrival.pingback_id)
 
             self.new_pingback_event.wait(timeout=1)
 
+    def failure_metric_watcher_thread(self, worker_thread_info):
+        """
+        Thread which is responsible for watching for test failures based on limits that are
+        exceeded.
+
+        Thesae checks were put into their own thread for 2 reasons:
```
drwill-ms, line 531: Misspelled 'These'

python/device/device.py
BertKleewein: We have a queue of outgoing messages and 10 threads pulling out of the queue and sending.  We do this because send_message blocks until the PUBACK comes back from the service (about 20 ms).  If we have a single thread pulling off the queue and sending messages one-at-a-time, performance drops through the floor.  If the connection drops, we just let the queue grow (no max size) until some of these threads get freed up by arriving PUBACKs.

---
In reply to: [512844308](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512844308) [](ancestors = 512844308)

drwill-ms: Ah, okay. Input queue, outgoing queue with 10 threads pulling off that queue. Sounds good.

python/device/device.py
BertKleewein: huh?

---
In reply to: [512842653](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512842653) [](ancestors = 512842653)

drwill-ms: the not teh

drwill-ms (approved):

python/device/device.py
BertKleewein: fixed


---
In reply to: [512867206](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512867206) [](ancestors = 512867206,512842653)

python/device/device.py
BertKleewein: unix epoch time stored as an integer.  I think I'll call this "timestamp" instead of "time," unless you have a better way to distinguish between a Time object and a unix epoch time integer.

---
In reply to: [512845578](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512845578) [](ancestors = 512845578)

BertKleewein: Actually, it's probably a float instead of an integer.

---
In reply to: [512874594](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512874594) [](ancestors = 512874594,512845578)

BertKleewein: But I'll change the name in a future commit since it will lead to merge conflicts for me right now.

---
In reply to: [512874969](https://github.com/Azure/iot-sdk-longhaul/pull/17#discussion_r512874969) [](ancestors = 512874969,512874594,512845578)

PR #18, opened 2020-10-28
rename fields and properties

BertKleewein:
This is the first part of a 3-part commit:
1 (this): Rename fields and properties for consistency.
2. Update pairing protocol to use telemetry and c2d.  This works much better than using twins. 
3. Add support for c2d testing

The ultimate purpose of these 3 commits is to test c2d.  
The pairing updates are necessary to cleanly start and stop the c2d code on the service side.  
The renames were done to make the code consistent and understandable in that context.

After this, I'm going to add code so the device can control attributes of the c2d messages such as when to start, when to stop, message cadence, and message size.

### DIFF
--- a/python/device/device.py
+++ b/python/device/device.py
@@ -19,9 +19,10 @@ from measurement import ThreadSafeCounter
 import azure_monitor
 from azure_monitor_metrics import MetricsReporter
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.WARNING)
 logging.getLogger("paho").setLevel(level=logging.INFO)
 logging.getLogger("thief").setLevel(level=logging.INFO)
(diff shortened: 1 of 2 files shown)


python/device/device.py
```diff
@@ -145,7 +144,7 @@ def __init__(self):
         self.received_pingback_list = []
         self.new_pingback_event = threading.Event()
         # for telemetry
-        self.telemetry_queue = queue.Queue()
+        self.outgoing_send_message_queue = queue.Queue()
```
drwill-ms, line 147: interesting that the previous change as the order of send_outgoing, but this one is outgoing_send.

BertKleewein: I'll take a look at that in my next PR.  That is interesting...consistency is very important in my world and I think I had a reason for this, but I'll make sure.

PR #19, opened 2020-10-29
update pingback code for upcoming pairing changes

BertKleewein:


python/device/device.py
```diff
@@ -402,49 +409,60 @@ def test_send_message_thread(self, worker_thread_info):
...
+
+                def on_pingback_received(pingback_id):
+                    self.metrics.send_message_count_received_by_service_app.increment()
+
+                with self.pingback_list_lock:
+                    self.pingback_wait_list[pingback_id] = MessageWaitingToArrive(
+                        callback=on_pingback_received,
+                        pingback_id=pingback_id,
+                        message=msg,
+                        send_time=time.time(),
```
drwill-ms, line 453: is this time UTC?

drwill-ms (approved):

python/device/device.py
BertKleewein: good reminder.  You asked that before and I promised a change.  time.time() returns unix epoch time as an integer (seconds since 1/1/1970).  I think calling it a timestamp might be better?  I'll change this in a future PR.  (I have cascading changes and my deltas get messed up if I'm not careful)

PR #20, opened 2020-10-29
update pairing logic to use messages instead of twin

BertKleewein:


python/device/device.py
```diff
@@ -263,99 +264,113 @@ def update_initial_reported_properties(self):
...
+                self.pairing_id = str(uuid.uuid4())
+                self.service_run_app_id = None
+                # set msg to None to trigger the block that sends the first pairingRequest message
+                msg = None
 
-        raise Exception("timeout waitig for service to set serviceAppRunId")
+            if not msg and currently_pairing:
+                if (
+                    time.time() - pairing_start_epochtime
+                ) > self.config.pairing_request_timeout_interval:
```
drwill-ms, line 320: I'm no python expert, but it looks like this ending parenthesis should be just before the colon

python/service/service.py
```diff
@@ -254,9 +213,31 @@ def handle_pairing_request_thread(self, worker_thread_info):
...
                 # TODO: remove items from list of no traffic for X minutes
 
-            # Step 4: sleep a while (but only if there's nothing in the queue)
-            if self.incoming_pairing_request_queue.empty():
-                time.sleep(10)
+    def unpair_device(self, device_id):
+        # TODO: this isn't very robust.  There should be some way to tell the device that we're done.
+        with self.pairing_list_lock:
+            if device_id in self.paired_devices:
+                logger.info("unpairing {}. Removing it from paired device list".format(device_id))
```
drwill-ms, line 220: Capitalize `Unpairing`?

python/service/service.py
```diff
@@ -254,9 +213,31 @@ def handle_pairing_request_thread(self, worker_thread_info):
...
+                if (
+                    pairing_id == self.paired_devices[device_id].pairing_id
+                    and service_app_run_id == run_id
+                ):
+                    if not self.paired_devices[device_id].pairing_complete:
+                        logger.info("Device {} has decided to pair with us".format(device_id))
+                        self.paired_devices[device_id].pairing_complete = True
+                else:
+                    logger.info(
+                        "Device {} has decided to pair with somebody else".format(device_id)
```
drwill-ms, line 238: If we know what instance it paired with, we could log it.

python/device/device.py
```diff
@@ -263,99 +264,113 @@ def update_initial_reported_properties(self):
...
+                    raise Exception(
+                        "No resopnse to pairing requests after trying for {} seconds".format(
+                            self.config.pairing_request_timeout_interval
+                        )
+                    )
 
-    def pairing_thread(self, worker_thread_info):
+                elif (
+                    time.time() - pairing_last_request_epochtime
+                ) > self.config.pairing_request_send_interval:
```
drwill-ms, line 329: same here

drwill-ms (approved):

python/device/device.py
drwill-ms: It looks like the if statement condition is meant to be comparing the difference in time to that constant. I just don't understand Python.

PR #21, opened 2020-11-03
add c2d testing

BertKleewein:


python/device/device.py
```diff
@@ -119,7 +131,16 @@ def __init__(self):
         self.send_message_unacked_failure_count = 200
 
         # How many send_message exceptions do we allow before we fail the test?
-        self.send_message_exception_failure_count = 5
+        self.send_message_exception_failure_count = 1
+
+        # How often do we want the service to send test C2D messages?
+        self.receive_message_interval = 2
```
drwill-ms, line 137: in seconds?

python/device/device.py
```diff
@@ -223,6 +265,7 @@ def get_longhaul_metrics(self):
 
         props = {
             "runStartUtc": self.metrics.run_start_utc.isoformat(),
+            "runCurrentTimeUtc": datetime.datetime.utcnow().isoformat(),
```
drwill-ms, line 268: start time of the run? latest update time?

python/device/device.py
```diff
@@ -518,9 +580,18 @@ def dispatch_incoming_message_thread(self, worker_thread_info):
                 if thief and thief["pairingId"] == self.pairing_id:
                     cmd = thief["cmd"]
                     if cmd == "pingbackResponse":
+                        print("received {} message with {}".format(cmd, thief["pingbacks"]))
```
drwill-ms, line 583: Why print sometimes, and logger.info other times?

python/device/device.py
```diff
@@ -119,7 +131,16 @@ def __init__(self):
...
+        self.send_message_exception_failure_count = 1
+
+        # How often do we want the service to send test C2D messages?
+        self.receive_message_interval = 2
+
+        # How big, in bytes, do we we want the random text size in test C2D messages to be
+        self.receive_message_filler_size = 16 * 1024
+
+        # How many missing C2D messages will cause the test to fail?
+        self.receive_message_missing_message_failure_count = 1
```
drwill-ms, line 143: I worry we're being too aggressive about killing a test for very strict conditions. I'd rather we try to go as long as we can (no catastrophic failures), and log the issues we see along the way. Then, from the logs, we can calculate the effective run time based on our (possibly changing) criteria. Even later, if we update our criteria, older runs that we didn't kill, can still be re-evaluated.

BertKleewein: good point.  You've said that before and I think I'm seeing the light.  I'll adjust.

python/device/device.py
BertKleewein: oops.  changing to logger.info.

python/device/device.py
BertKleewein: renaming

drwill-ms (approved):

PR #23, opened 2020-11-09
make python docker scripts generic

BertKleewein:


python/device/out_of_order_message_tracker.py
```diff
@@ -41,20 +41,23 @@ def add_message(self, index):
...
                 self.max_resolved_index = index - 1
 
-            while len(self.unresolved_indices):
-                if (self.max_resolved_index + 1) in self.unresolved_indices:
-                    self.max_resolved_index += 1
-                    self.unresolved_indices.remove(self.max_resolved_index)
-                else:
-                    break
+            # since we're qos=1, we can get a c2d more than once
+            # This is why we comapre to max_resolved_index
```
drwill-ms, line 51: `compare`

scripts/run-container.sh
```diff
@@ -0,0 +1,86 @@
...
+function usage {
+    echo "USAGE: ${0} [--platform platform] --langauge language_short_name --library library [--source library_source] --version library_version --pool service_pool [--device_id device_id] [--tag extra_tag]"
+    echo "  ex: ${0} --language py37 --library device --source pypi --version 2.3.0 --pool pool_1"
+    exit 1
+}
+
+source ${script_dir}/_parse_args $*
+
+if [ "${LANGUAGE_SHORT_NAME}" == "" ]; then
+    echo "ERROR: Language is required"
```
drwill-ms, line 16: Why not match the exact environment variable name?

drwill-ms (approved):
Most of this look like Greek to me, but if you run it and it works then it must be pretty good. :)

PR #24, opened 2020-11-16
Docs and scripts

BertKleewein:


docs/azure-monitor-logs.md
```diff
@@ -0,0 +1,57 @@
+# Looking at test logs and metrics in Azure Monitor
+
+1. Open the `thief-ai` resource in the Azure portal
+2. On the left pane, click "Logs" in the "Monitoring" section
```
drwill-ms, line 4: Did you know in markdown you can number every bullet `1.` and it will automatically increment when rendered?

docs/launching-remote-tests.md
```diff
@@ -0,0 +1,99 @@
...
+We currently support the following sources:
+
+| langauge | source | language default? | library_version format |
+| - | - | - | - |
+| Python | pypi | yes |  package version.  e.g. "2.4.0" |
+
+The libraries are built using `scripts/build_image.sh`.
+
+```
+USAGE: ./build-image.sh [--platform platform] --langauge language_short_name --library library [--source library_source] --version library_version [--tag extra_tag]
```
drwill-ms, line 32: What if I want to use .NET service client and NodeJS device client?

drwill-ms (approved):

docs/azure-monitor-logs.md
BertKleewein: Yes, I did know that, but that feels like "more than formatting" to me, so I generally don't use it.

docs/launching-remote-tests.md
BertKleewein: that's possible.  build-image.sh builds either the device or the service app, so you would just adjust the parameters to build/run the mix.

docs/azure-monitor-logs.md
drwill-ms: ah, I personally like it because adding or editing steps doesn't create a tiresome task of updating all the numbers, or forgetting and have them not make sense.

PR #25, opened 2020-11-19
add device twin reported property testing

BertKleewein:


python/common/app_base.py
```diff
@@ -44,6 +46,24 @@ def disconnect(self):
...
 
+    def pause_all_threads(self):
+        """
+        Pause all threads.  Used to stop all background threads while debugging
+        """
+        self.paused = True
+
+    def unpause_all_threads(self):
+        """
+        Unpause all threads.  Used when debugging is finished./
```
drwill-ms, line 57: Spurious character: `/`

python/device/device.py
```diff
@@ -162,10 +191,12 @@ def __init__(self):
         self.pingback_wait_list = {}
         self.incoming_pingback_response_queue = queue.Queue()
         # for telemetry
-        self.outgoing_send_message_queue = queue.Queue()
+        self.outgoing_test_message_queue = queue.Queue()
```
drwill-ms, line 194: What is a `test message`?

python/device/device.py
```diff
@@ -308,8 +373,59 @@ def update_initial_reported_properties(self):
...
+        if self.service_app_run_id:
+            props["thief"]["serviceAppRunId"] = self.service_app_run_id
+        props["thief"]["pairingId"] = self.pairing_id
+
+        # This function only creates the message.  The caller needs to queue it up for sending.
+        msg = Message(json.dumps(props))
+        msg.content_type = "application/json"
+        msg.content_encoding = "utf-8"
+
+        msg.custom_properties["eventDateTimeUtc"] = datetime.datetime.now(
```
drwill-ms, line 395: `datetime.datetime`?

python/device/device.py
```diff
@@ -308,8 +373,59 @@ def update_initial_reported_properties(self):
...
+            datetime.timezone.utc
+        ).isoformat()
+
+        return msg
+
+    def create_message_from_dict_with_pingback(
+        self, props, on_pingback_received, pingback_type, user_data=None
+    ):
+        """
+        helper function to crate a message from a dict and add pingback
```
drwill-ms, line 405: misspelled create

python/device/device.py
```diff
@@ -646,37 +756,100 @@ def check_for_failure_thread(self, worker_thread_info):
...
+                                pingback_wait_info.pingback_id,
+                                (now - pingback_wait_info.send_epochtime),
+                                self.config.reported_properties_update_interval_in_seconds,
+                            )
+                        )
+                        reported_properties_add_failure_count += 1
+
+                    elif (
+                        (
+                            pingback_wait_info.pingback_type
```
drwill-ms, line 801: Is this newline just because of line length?

python/device/device.py
```diff
@@ -758,10 +929,117 @@ def handle_test_c2d_messages_thread(self, worker_thread_info):
...
+            try:
+                msg = self.outgoing_control_message_queue.get(timeout=1)
+            except queue.Empty:
+                msg = None
+            if msg:
+                try:
+                    self.client.send_message(msg)
+                except Exception as e:
+                    logger.warning(
+                        "send_message raised {} sending a control messsage.  Ignoring".format(
```
drwill-ms, line 955: Is there no retry if the message fails to send?

I mean, if the device were disconnected for a bit, would this message be lost?

In the .NET impl, I keep retrying the top message until it goes through.

drwill-ms (approved):

python/device/device.py
BertKleewein: We can talk about this when we meet.  I've divided device-to-cloud messages into 2 categories:
* test messages, where we send at a regular rate and verify receipt.
* control messages, where we try to influence the behaviour of the service app.

drwill-ms: My point is, as a code reader, the word `test` is so ambiguous I have no idea what meaning you intend. I think it is poorly named.

python/device/device.py
BertKleewein: yes, I think so, but formatting decisions are made by the pretty-printer that we use.  I write code that is syntactically correct, and the pretty-printer makes it look consistent.  Sometimes it comes out ugly, but the rules are simple and consistent and mostly accepted as "the right way" by the community.

python/device/device.py
BertKleewein: There is a retry, but there could still be a fatal (non-retryable) exception.  Unfortunately, our current Python SDK will hold the message until it sends, with no timeout and no exponential backoff.

python/device/device.py
drwill-ms: Weird Python naming

python/device/device.py
drwill-ms: Maybe a code comment would be helpful, given each SDK has its own behavior in cases like this.

PR #26, opened 2020-12-07
support multiple deployments, update service logging to include devic…

BertKleewein:
* support multiple deployments, including a script to create Azure resources.  Will investigate bicep tool as a possible later change.
* update service logging to include device details.
* more documentation.
* update report schema with sub objects and rename pairingId to sessionId.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1,0 +1,71 @@
+# THIEF environment variables
+
+verion .1
+
+## Coulmns
+
+| column name | meaning |
+| - | - |
+| variable name | name of the environment variable |
+| service | variable is available to service apps when running inside Azure container instances |
+| device | variable is available to device apps when running inside Azure container instances |
+| developer | variable is available to developers via fetch-secrets.sh |
+| source | where the value comes from.  Either 'keyvault' for variables that come from keyvault, or 'manual' for variables that are manually set via some other route (script) |
+| format | format of the variable |
+| meaning | what the variable represents |
+
+## Keyvault conventions
+Since keyvault doesn't support underscores in secret names, keyvault secrets are names the same as the variable names, with underscores (\_) replaced with dashes (-)
+
+e.g. The environment variable `THIEF_SUBSCRIPTION_ID` is stored as a keyvault seret named `THIEF-SUBSCRIPTION-ID`
+
+## Azure resources
+| variable name | service | device | developer | source | format | meaning |
+| - | - | - | - | - | - | - |
+| THIEF\_SUBSCRIPTION\_ID | - | - | X | keyvault | GUID | Azure subscription ID holding all THIEF resources | 
+| THIEF\_ACTIVE\_DIRECTORY\_TENANT | - | - | X | keyvault | hostname | Azure active directory tenant.  Uesd to consturct portal URLs. |
+| THIEF\_RESOURCE\_GROUP | - | - | X | keyvault | string | Name of resource group containing all THIEF resources |
+| THIEF\_RUNS\_RESOURCE\_GROUP | - | - | X | keyvault | string |  Name of resource group containing all THIEF container instances |
+| THIEF\_USER\_RESOURCE\_ID | - | - | X | keyvault | string | Azure resource ID for the managed identity that container instances run uner.  Used to get access to keyvault secrets from inside container instances |
+
+## App Insights
+| variable name | service | device | developer | source | format | meaning |
+| - | - | - | - | - | - | - |
+| THIEF\_APP\_INSIGHTS\_CONNECTION\_STRING | X | X | X | keyvault | connection string | Connection string used to push secrets into App Insights.  Starts wiuth `InstrumentationKey=`. |
+| THIEF\_APP\_INSIGHTS\_NAME | - | - | X | keyvault | string | name of App Insights resouce |
+
+## IoTHub Service
+| variable name | service | device | developer | source | format | meaning |
+| - | - | - | - | - | - | - |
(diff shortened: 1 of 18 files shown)


docs/environment-variables.md
```diff
@@ -0,0 +1,71 @@
+# THIEF environment variables
+
+verion .1
+
+## Coulmns
```
drwill-ms, line 5: misspelled

docs/names-and-ids.md
```diff
@@ -0,0 +1,43 @@
...
+| Identifier | format | Meaning |
+| - | - | - |
+| servicePool | string | Name of pool that a service app belongs to. |
+| requestedServicePool | string | Name of pool that a device app would like to pair with. |
+| runId | guid | Guid representing a running executable. |
+| serviceAppRunId | guid | Guid representing a running _service_ executable. |
+| sessionId | guid | Guid representing the pairing between a device app and a service app. |
+| pingbackId | guid | Guid used to represent the ACK of a thief operation. |
+
+## servicePool notes
```
drwill-ms, line 16: `servicePool` should either be capitalized or, if it is a variable, should be wrapped in backticks

python/device/device.py
```diff
@@ -1092,6 +1128,9 @@ def main(self):
 
         logger.info("Exiting main at {}".format(datetime.datetime.utcnow()))
 
+        # 60 seconds to let app insights data flush
+        time.sleep(60)
```
drwill-ms, line 1132: In .NET we have a way to force the AppInsights SDK to flush. No such functionality in the Python version?

If you have to wait 60 seconds, you probably should print something to the console so anyone running it locally doesn't think it is hanging.

python/service/service.py
```diff
@@ -143,7 +154,10 @@ def __init__(self):
 
     def add_device_to_pairing_list(self, event):
         device_id = get_device_id_from_event(event)
-        logger.info("pairing request received for device {}. Queueing.".format(device_id))
+        logger.info(
+            "pairing request received for device {}. Queueing.".format(device_id),
```
drwill-ms, line 158: Proper capitalization for log messages?

docs/names-and-ids.md
```diff
@@ -0,0 +1,43 @@
...
+
+There are a number of names and IDs used in THIEF.  This table describes some of them:
+
+| Identifier | format | Meaning |
+| - | - | - |
+| servicePool | string | Name of pool that a service app belongs to. |
+| requestedServicePool | string | Name of pool that a device app would like to pair with. |
+| runId | guid | Guid representing a running executable. |
+| serviceAppRunId | guid | Guid representing a running _service_ executable. |
+| sessionId | guid | Guid representing the pairing between a device app and a service app. |
```
drwill-ms, line 13: How about `pairingId` then?

docs/names-and-ids.md
```diff
@@ -0,0 +1,43 @@
...
+
+version .1
+
+There are a number of names and IDs used in THIEF.  This table describes some of them:
+
+| Identifier | format | Meaning |
+| - | - | - |
+| servicePool | string | Name of pool that a service app belongs to. |
+| requestedServicePool | string | Name of pool that a device app would like to pair with. |
+| runId | guid | Guid representing a running executable. |
```
drwill-ms, line 11: Is this device-specific?

docs/names-and-ids.md
```diff
@@ -0,0 +1,43 @@
...
+* Since service fallback isn't implemented, there currently no advantage to having more than one service app in any given pool.
+
+## requestedServicePool notes
+* A device specifies the nam of a service pool that it would like to pair with.
+* The device _must_ specify a requested service pool.
+* Any service app in the requested pool can pair with the device app.
+* A service app cannot pair with a device app that is requesting a different pool.
+* If no service apps are available in the device's requested pool, the pairing fails. 
+
+## runId notes
```
drwill-ms, line 29: We probably should have the device include the device runId in it's messages and twin, so when the service logs, it includes that runId in custom properties. Same for service including its runId in communications so the device can log that it its custom properties.

This way we can find all the associated logs for a device's run (including the relevant service logs).

Thoughts?

drwill-ms (approved):

docs/names-and-ids.md
BertKleewein: Yes, we should do that.

docs/names-and-ids.md
BertKleewein: Yes, but the names are still confusing.  I think I'm going to collapse runId and sessionId/pairingId into a single value -- the reason I made them separate is no longer valid.    Since we need an ID for the device app and  different ID for the service app, maybe "serviceAppSessionId" and "deviceAppSessionId".  What do you think of that?

python/device/device.py
BertKleewein: I'm going to go back and check.  This was a dirty fix but an explicit flush is better.

docs/names-and-ids.md
BertKleewein: it was pairingId, but that wasn't quite right.  See my comment on collapsing runId and sessionId into a single value.  I think that will be less confusing.

drwill-ms: Well, presumably a device starts up, and gets a pairing. If it keeps that pairing for the duration of the lifetime, then the device run Id could serve the same purpose. However, if it could lose a service and pick up another then you'd want another value: pairing Id.

docs/names-and-ids.md
drwill-ms: I'm fine with that, but I'd use `serviceRunId` and `deviceRunId` for brevity, as I think it is sufficiently descriptive

PR #28, opened 2020-12-14
use desired and reported properties for pairing and test control

BertKleewein:


python/device/device.py
```diff
@@ -514,34 +482,94 @@ def pairing_thread(self, worker_thread_info):
                 elif (
                     time.time() - pairing_last_request_epochtime
                 ) > self.config.pairing_request_send_interval_in_seconds:
+                    logger.info("pairing response timeout.  Requesting again")
```
drwill-ms, line 485: nit: Mixed casing styles for these two sentences.

drwill-ms: And elsewhere

python/device/device.py
```diff
@@ -514,34 +482,94 @@ def pairing_thread(self, worker_thread_info):
...
-                if currently_pairing:
-                    thief = json.loads(msg.data.decode()).get("thief")
-                    if thief["cmd"] == "pairingResponse":
-                        service_app_run_id = thief["serviceAppRunId"]
-                        logger.info(
-                            "Service app {} claimed this device instance".format(service_app_run_id)
+            elif msg and isinstance(msg, dict):
+                # is msg is a dict, that means we have a desired property change.  A service
+                # might be trying to pair with us.
+                logger.info("received pairing desired props: {}".format(pprint.pformat(msg)))
```
drwill-ms, line 519: Why not base the pairing request off a message property?

python/device/device.py
```diff
@@ -712,27 +764,31 @@ def dispatch_incoming_message_thread(self, worker_thread_info):
...
+                        # If this is a pingback response, we put it into `incoming_pingback_response_queue`
+                        # for another thread to handle.
                         logger.info("received {} message with {}".format(cmd, thief["pingbacks"]))
                         self.incoming_pingback_response_queue.put(msg)
-                    elif cmd == "pairingResponse":
-                        logger.info("received {} message".format(cmd))
-                        self.incoming_pairing_message_queue.put(msg)
+
                     elif cmd == "testC2d":
+                        # If thie is a test C2D messages, we put it into `incoming_test_c2d_message_queue`
```
drwill-ms, line 783: misspelled 'this'

python/device/device.py
```diff
@@ -712,27 +764,31 @@ def dispatch_incoming_message_thread(self, worker_thread_info):
...
 
             if msg:
                 obj = json.loads(msg.data.decode())
                 thief = obj.get("thief")
 
                 if thief and thief["pairingId"] == self.pairing_id:
+                    # We only inspect messages that have `thief/pairingId` set to our `pairingId` value.
                     cmd = thief["cmd"]
                     if cmd == "pingbackResponse":
+                        # If this is a pingback response, we put it into `incoming_pingback_response_queue`
```
drwill-ms, line 777: What is pingback?

scripts/create-thief-resources.sh
```diff
@@ -22,7 +22,7 @@ export THIEF_KEYVAULT_NAME=${THIEF_PREFIX}-thief-kv
 ##############################
 # add the app insights az extension
 ##############################
-echo az extension add --name application-insights
+az extension add --name application-insights
```
drwill-ms, line 25: Consider making an ARM template instead so it works across languages/platforms. Also, in my experience the CLI breaks you fairly often.

Also try using Bicep. See my team's 1N under Testing | Bicep for some instructions.

drwill-ms (approved):

python/device/device.py
BertKleewein: I'm documenting this now.  The pairing process is a handshake with 4 steps, all done with reported and desired properties:                                

1. A device app says "I need a partner" by setting it's reported `serviceRunId` to `null`.                                     
2. A service app says "I'm available" by setting the devices desired `serviceRunId` to the service app's `runId` value.        
3. The device app says "I choose you" by setting it's reported `serviceRunId` to the `runId` of the chosen service             
4. The service app finalizes the pairing by setting the devices desired `acceptedPairing` value.                               
                                                                                                                               
Once `acceptedPairing` is set, the pairing is complete and test operation can begin.

scripts/create-thief-resources.sh
BertKleewein: absolutely.  I just haven't had time to do this yet.

python/device/device.py
BertKleewein: This is the word I've been using for the service app acknowledging something.  telemetry and reported property changes both generate this ack.  I don't think it's a great name but I'm not sure what would be better.  Do you have any suggestions?

python/device/device.py
drwill-ms: 1-3 make sense to me. Why is 4 necessary?

python/device/device.py
drwill-ms: ServiceAck?

python/device/device.py
BertKleewein: Because of network latency and out-of-order delivery.  
The device app needs to know when it can start "regular operation" 
After step 3, the device app knows that it sent the "pairing is complete" message, but it doesn't know that the service app received it, and it doesn't know if the service app is "ready for regular traffic".

drwill-ms: Hmm, okay. I'll buy that.

Back to my original question, why do we figure out the type of message based on the type (dictionary)? That seems like a weak connection.

BertKleewein: Ah, I misunderstood your question.  You're right.  This was not done well.  I'm going to revisit this in a later commit.

PR #32, opened 2021-01-05
move strings into constants

BertKleewein:


python/common/thief_constants.py
```diff
@@ -3,7 +3,91 @@
...
+    JSON_TYPE_AND_ENCODING = {
+        "contentType": "application/json",
+        "contentEncoding": "utf-8",
+    }
+    PROPERTIES = "properties"
+    REPORTED = "reported"
+
+
+class Fields(object):
+    class Deprecated(object):
```
BertKleewein, line 18: >Deprecated [](start = 10, length = 10)

I'm going to remove the functionality around these strings, but I wanted to do that as a separate PR.

python/device/device.py
```diff
@@ -310,7 +314,7 @@ def get_session_metrics(self):
             "runTime": str(self.metrics.run_time),
             "runState": str(self.metrics.run_state),
             "exitReason": self.metrics.exit_reason,
-            "pairingId": self.pairing_id,
+            Fields.Deprecated.PAIRING_ID: self.pairing_id,
```
BertKleewein, line 317: I'm going to move other strings into thief_constants.py, but not in this change.

python/common/thief_constants.py
```diff
@@ -3,7 +3,91 @@
...
 
-class ServiceAckType(object):
-    TELEMETRY_SERVICE_ACK = "telemetry"
-    ADD_REPORTED_PROPERTY_SERVICE_ACK = "add_reported"
-    REMOVE_REPORTED_PROPERTY_SERVICE_ACK = "remove_reported"
+class Const(object):
+    JSON_CONTENT_TYPE = "application/json"
+    JSON_CONTENT_ENCODING = "utf-8"
+    JSON_TYPE_AND_ENCODING = {
+        "contentType": "application/json",
```
drwill-ms, line 10: Can this value reference JSON_CONTENT_TYPE?

python/common/thief_constants.py
```diff
@@ -3,7 +3,91 @@
...
+        CONFIG = "config"
+
+        PAIRING = "pairing"
+
+        class Pairing(object):
+            REQUESTED_SERVICE_POOL = "requestedServicePool"
+            SERVICE_RUN_ID = "serviceRunId"
+            DEVICE_RUN_ID = "deviceRunId"
+
+        TEST_CONTENT = "testContent"
```
drwill-ms, line 46: Is this a stand-alone constant? I.E. not in a class?

python/device/device.py
```diff
@@ -635,7 +641,7 @@ def send_metrics_to_azure_monitor(self, props):
         """
         # we don't record session_metrics to azure monitor
 
-        system_health_metrics = props["systemHealthMetrics"]
+        system_health_metrics = props[Fields.Reported.SYSTEM_HEALTH_METRICS]
         self.reporter.set_process_cpu_percent(system_health_metrics["processCpuPercent"])
```
drwill-ms, line 645: TODO: "processCpuPercent" and the others below?

drwill-ms (approved):

python/common/thief_constants.py
BertKleewein: not cleanly.  There's a few different ways to hack it, but they would only add confusion.

PR #34, opened 2021-01-05
schema updates based on previous PR feedback

BertKleewein:
Various minor schema changes:
* removed random "fillter" from c2d messages
* update doc for sdkVersion field
* flattened serviceAcks array to be an array of guids instead of an array of objects that contain guids
* removed unused "firstMessage" field from c2d
* updated osRelease implementation to follow user agent convention


docs/test-content.md
```diff
@@ -16,12 +16,8 @@ Because c2d is a fairly expensive operation, the service app gathers `serviceAck
       "serviceRunId": "ca3ceadb-ac57-4cca-8f9b-c5d1e4dc189d",
       "pairingId": "3887e97e-6f06-484b-a5bf-1753913866d3",
       "serviceAcks": [
```
drwill-ms, line 18: `serviceAckIds`?

drwill-ms (approved):

docs/test-content.md
BertKleewein: I'm going to keep it as serviceAcks for now.  There's a chance that I'll have to expand it to include more data per ack (timestamps specifically).

PR #36, opened 2021-01-07
Log a metric for message delay

drwill-ms:


net/device/IotHub.cs
```diff
@@ -261,6 +262,7 @@ private void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionS
                     {
                         { "DisconnectedStatus", _disconnectedStatus.ToString() },
                         { "DisconnectedReason", _disconnectedReason.ToString() },
+                        { "ConnectionStatusChangeCount", _connectionStatusChangeCount.ToString() },
```
drwill-ms, line 265: I log this because the order of traces is not guaranteed, and I want to know the order of dis/connection events as seen by the app.

net/device/IotHub.cs
```diff
@@ -123,6 +123,7 @@ public async Task RunAsync(CancellationToken ct)
 
                         ++_totalMessagesSent;
                         _logger.Metric("TotalMessagesSent", _totalMessagesSent);
+                        _logger.Metric("MessageDelayMinutes", (DateTime.UtcNow - pendingMessage.CreationTimeUtc).TotalMinutes);
```
drwill-ms, line 126: I haven't synced with the spec yet, but I added this for my interim runs so I can see how long messages are delayed.

net/device/IotHub.cs
```diff
@@ -11,6 +11,7 @@
 using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
+using static Microsoft.Azure.Iot.Thief.Device.LoggingConstants;
```
BertKleewein, line 14: This is interesting.  How is a static class different from a namespace, and did you have a reason for choosing it here?

BertKleewein (approved):
:shipit:

net/device/IotHub.cs
drwill-ms: A static class is just one you can't instantiate. Referencing it in a using statement like this simply means this code can refer to properties and methods in that class without using the class name.

BertKleewein: Right, I got that part.  I was wondering why you didn't make a namespace called "Microsoft.Azure.Iot.Thief.Device.LoggingConstants" and put the strings into that namespace.  I just remembered that you can't do this.  (https://stackoverflow.com/questions/2818333/namespace-constant-in-c-sharp)

drwill-ms: Yep, in .NET everything has to be in a class.

PR #38, opened 2021-01-11
Use PascalCase for azure monitor metric names

BertKleewein:


python/device/device.py
```diff
@@ -610,31 +583,10 @@ def send_metrics_to_azure_monitor(self, props):
...
-        self.reporter.set_send_message_count_in_backlog(test_metrics["sendMessageCountInBacklog"])
-        self.reporter.set_send_message_count_unacked(test_metrics["sendMessageCountUnacked"])
-        self.reporter.set_send_message_count_not_received_by_service_app(
-            test_metrics["sendMessageCountNotReceivedByServiceApp"]
-        )
-        self.reporter.set_receive_message_count_received(
-            test_metrics["receiveMessageCountReceived"]
-        )
-        self.reporter.set_receive_message_count_missing(test_metrics["receiveMessageCountMissing"])
+        # we don't record session_metrics to azure monitor because they're all about time and i
```
drwill-ms, line 586: Can you re-read this sentence and see if it makes sense to you? It seems the end of this line doesn't continue with the next in a way that makes sense to me.

drwill-ms (approved):

python/device/device.py
BertKleewein: I remember feeling awkward when I was writing that.  I'll try again :)

PR #42, opened 2021-01-13
simplify metrics to make failure count more apparent

BertKleewein:


python/common/thief_constants.py
```diff
@@ -93,3 +93,52 @@ class Events(object):
     SENDING_PAIRING_REQUEST = "SendingPairingRequest"
     RECEIVED_PAIRING_RESPONSE = "ReceivedPairingResponse"
     PAIRING_COMPLETE = "PairingComplete"
+
+
+class MetricNames(object):
+    # ---------------------
+    # System Health metrics
+    # ---------------------
+    PROCESS_CPU_PERCENT = "processCpuPercent"
```
drwill-ms, line 102: These should be PascalCased per the spec, right?

drwill-ms (approved):

python/common/thief_constants.py
BertKleewein: Yes.  Metrics are PascalCased when pushed to Azure Monitor and camelCased when included in JSON.

PR #45, opened 2021-01-19
Reduce amount of log traces being pushed to Azure Monitor.

BertKleewein:
When I added latency metrics to Azure Monitor, it started dropping log tracing.  To fix this, I'm going to limit the quantity of traces that go to Azure Monitor and add a way to capture the log stream, possibly pushing it to an Azure file share.

### DIFF
--- a/python/common/thief_constants.py
+++ b/python/common/thief_constants.py
@@ -254,3 +254,22 @@ class DeviceSettings(object):
     REPORTED_PROPERTIES_UPDATE_INTERVAL_IN_SECONDS = "reportedPropertiesUpdateIntervalInSeconds"
     # How many reported property patches are allowed to fail before we fail the test
     REPORTED_PROPERTIES_UPDATE_ALLOWED_FAILURE_COUNT = "reportedPropertiesUpdateAllowedFailureCount"
+
+
+class CustomDimensionNames(object):
+    """
+    Names of cusomDimension fields pushed to Azure Monitor
+    """
+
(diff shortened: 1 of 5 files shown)


python/common/thief_constants.py
```diff
@@ -254,3 +254,22 @@ class DeviceSettings(object):
     REPORTED_PROPERTIES_UPDATE_INTERVAL_IN_SECONDS = "reportedPropertiesUpdateIntervalInSeconds"
     # How many reported property patches are allowed to fail before we fail the test
     REPORTED_PROPERTIES_UPDATE_ALLOWED_FAILURE_COUNT = "reportedPropertiesUpdateAllowedFailureCount"
+
+
+class CustomDimensionNames(object):
+    """
+    Names of cusomDimension fields pushed to Azure Monitor
```
drwill-ms, line 261: misspelled custom

python/common/thief_constants.py
```diff
@@ -254,3 +254,22 @@ class DeviceSettings(object):
...
+    """
+
+    OS_TYPE = "osType"
+    SDK_LANGUAGE = "sdkLanguage"
+    SDK_LANGUAGE_VERSION = "sdkLanguageVersion"
+    SDK_VERSION = "sdkVersion"
+
+    SERVICE_INSTANCE = "serviceIntsance"
```
drwill-ms, line 269: misspelled instance

drwill-ms (approved):

PR #46, opened 2021-01-26
use bicep for deployment, seperate shared and instance resources

BertKleewein:


scripts/thief.json
```diff
@@ -0,0 +1,473 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
```
BertKleewein, line 2: This file is generated from thief.bicep and does not need to be reviewed

docs/azure-resources.md
```diff
@@ -0,0 +1,86 @@
...
+This is necessary to support developers running under their own subscriptions while still reporting to the global stores.
+
+## Shared THIEF resources
+
+All of these resources are under the subscription `THIEF_SHARED_SUBSCRIPTION_ID` in the `THIEF_SHARED_RESOURCE_GROUP` resource group.
+
+`THIEF_SHARED_KEYVAULT_NAME` has a keyvault with secrets for all shared resources.
+This kevault is used to seed the instance keyvault and many of the secrets in this keyvault are duplicated in the instance keyvault
+
+`THIEF_APP_INSIGHTS_CONNECTION_STRING` is the entrypoint for the shared App Insights intance.
```
drwill-ms, line 25: In .NET I'm just using the AppInsights Key, not a connection string. What options do you have in Python?

docs/azure-resources.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+2. Use `az account set --subsciption <foo>` to set the subsciption you want to use for your instance resources.  Use `az account list` if you need help finding the subscription id.
+```
+(longhaul) bertk@bertk-hp:~/repos$ az account set --subscription "<REDACTED>"
+```
+
+3. Call `scripts/create-thief-instance.sh`, passing in a resource prefix (such as your email address) to create the thief resources.
+```
+(longhaul) bertk@bertk-hp:~/repos/longhaul$ ./scripts/create-thief-instance.sh bertk7 --location westus2
```
drwill-ms, line 62: I'm thinking this won't work for me on Windows. Thoughts on cross-platform support?

scripts/thief.bicep
```diff
@@ -0,0 +1,346 @@
...
+    accessPolicies: [
+      {
+        tenantId: subscription().tenantId
+        objectId: thief_container_identity.properties.principalId
+        permissions: {
+          secrets: [
+            'get'
+          ]
+        }
+      }
```
drwill-ms, line 99: should there be a comma at the end of this line? These look like objects in an array.

scripts/thief.bicep
```diff
@@ -0,0 +1,346 @@
...
+      {
+        tenantId: subscription().tenantId
+        objectId: user_principal_id
+        permissions: {
+          keys: [
+            'get'
```
vinagesh, line 105: you can just use 'All'

scripts/thief.bicep
vinagesh: I've not see a usage of comma even for arrays in bicep

scripts/thief.bicep
```diff
@@ -0,0 +1,346 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license information.
+
+param location string
```
vinagesh, line 5: Adding atleast metadata for params is good. Defaults can be optional.

param location string {
  default: '${resourceGroup().location'
  metadata: {
    description: 'The location of the resource.'
  }
}

scripts/thief.bicep
```diff
@@ -0,0 +1,346 @@
...
+param prefix string
+param thief_runs_resource_group string
+param thief_app_insights_connection_string string
+param thief_container_registry_host string
+param thief_container_registry_password string
+param thief_container_registry_shortname string
+param thief_container_registry_user string
+param user_principal_id string
+param thief_shared_subscription_id string
+param thief_shared_keyvault_name string
```
vinagesh, line 15: Would be good to add min and max length for KV resource
Ex:

param KeyVaultName string {
  default: '${resourceGroup().name}-kv'
  minLength: 3
  maxLength: 24
  metadata: {
    description: 'The name of the key vault for xxx'
  }
}

docs/azure-resources.md
BertKleewein: I think it's the same thing -- `InstrumentationKey=<guid>`.  The Python `AzureEventHandler` class, etc, call the _parameter_ `connection_string`, so I carried that forward.

docs/azure-resources.md
BertKleewein: There's 2 issues here.  `create-thief-resources.sh` is one, but it only gets called once if you need to run against a private hub.  The other issue is `fetch-secrets.sh` which you use to get secrets from the vault and into the environment.  We definitely want a Windows version of fetch-secrets.sh.  I'm not sure about create-thief-instance.

docs/azure-resources.md
drwill-ms: So not a connection string??

docs/azure-resources.md
drwill-ms: Are you saying I need to run Linux to participate in Thief?

BertKleewein: Right now, yes, Linux is required.  Sort of.    

I would say that you need to run Linux to use some of the more automated aspects.  

I agree that this is a problem that we need to solve.  Bicep is a good start, but there are still things it doesn't do.  We still need scripts for:
1. creating DPS groups and enrollments.
2. fetching secrets (currently into environment variables)
3. building and launching containers.

We can make all of these things work cross-platform.   We just have to decide how and what priority.

(Side note: Windows container support is getting better, but still lags behind Linux.  We may need a different environment (like VMs) when we want automated tests under Windows).

docs/azure-resources.md
BertKleewein: Looking at the AI portal, it appears to be a partial connection string (InstrumentationKey only, IngestionEndpoint not included).  It sounds like you're saying that a raw Instrumentation Key would be more useful.  Is that right?  If so, I have no problem changing it to a key and renaming it.

scripts/thief.bicep
BertKleewein: nice!  thank you!

docs/azure-resources.md
drwill-ms: The raw AI key is all I need for .NET, yes.

docs/azure-resources.md
drwill-ms: I thought PowerShell was a cross platform option. Would that work for both Linux and Windows?

drwill-ms: To be clear, I'm talking about the developer environment, which is where I assume this script is being run.

BertKleewein: Yes, the developer environment.  PowerShell is an option.  There's also bash on Windows, which I assume goes through WSL.

drwill-ms: WSL is "sorta" cross platform, but not as smooth an experience as PowerShell (I'd assume, but not having used PowerShell on Linux I can't say it is great).

I'll accept WSL for now, but perhaps we can look at streamlining this experience.

drwill-ms (approved):

docs/azure-resources.md
BertKleewein: fixed in #48

scripts/thief.bicep
BertKleewein: Agree that this could be useful, but I don't want to introduce any limits at this point.

drwill-ms: The good reason to do it is it'll fail immediately if the parameters aren't right. These are known limits in Azure for KV.

If you don't check it, and someone passes a bad value, it'll take longer because it must deploy and fail. If it fails, you may have some resources done, but not others, and that requires clean up.

BertKleewein: OK, you talked me into it.

PR #52, opened 2021-02-01
add UML diagrams for test flow

BertKleewein:


docs/asset-source/test-telemetry.puml
```diff
@@ -0,0 +1,17 @@
+@startuml
+participant "Device App" as D
+participant "Service App" as S
+
+title **Telemetry Testing**
+
+D -> S: SendTelemetry(\n\tcmd="serviceAckRequest", \n\tserviceInstanceId=<instanceId>, \n\tserviceAckId=<ackId_1>)
```
abhipsaMisra, line 7: q - is all of this in the payload `(cmd="serviceAckRequest", serviceInstanceId=<instanceId>, serviceAckId=<ackId_1>)`? Is it a json?

docs/asset-source/test-c2d.puml
```diff
@@ -0,0 +1,18 @@
...
+title **C2D testing**
+
+D -> S: SetReportedProperties({\n\ttestControl: {\n\t\tc2d: {\n\t\t\tsend: true\n\t\t\tmessageIntervalInSeconds: x\n}}})
+note left: Device asks service to start sending\nc2d messages, one every x seconds
+
+S -> D: SendC2d({\n\tcmd: "testC2d"\n\ttestC2dMessageIndex: 1\n})
+note right: Service begins sending c2d,\none every x seconds\nwith increasing testC2dMessageIndex values
+
+S -> D: SendC2d({\n\tcmd: "testC2d"\n\ttestC2dMessageIndex: 2\n})
+note left: No device response is necessary
```
abhipsaMisra, line 14: Q - any reason why we don't validate the actual receipt of the c2d message by the device? The reason I ask is, we do get issues where customer's devices show unreliable behavior while executing the c2d flow; so it might be worth adding it to our tests

docs/asset-source/test-telemetry.puml
BertKleewein: It is JSON in the payload.  I think I'll update this to be explicit.

docs/asset-source/test-c2d.puml
BertKleewein: Good point.  The detail that is missing is that the device is still validating reception of c2d by watching the testC2dMessageIndex values.  Any missing index values are logged and tracked, with a failure after X messages are lost. 

There is also the issue of, I think it's called "settlement", where AMQP lets the device acknowledge, reject, or, (I think) ignore a message, and that gets reported back to the service.   I haven't dealt with this since Python is MQTT only.  Do you think we need to validate this?

abhipsaMisra: I don't think validating the Reject and Abandon functionalities of AMQP would bring much value in long-haul scenarios; we have e2e tests validating that flow for a single device - single message, and that should be sufficient from the sdk's perspective. The only issue I see arising from abandoning messages is that we might exceed the queue size, but that to me is something the service will need to handle/ test (also, subsequent error after this would be encountered at the service client side, and not at the device side).
I think validating the receive -> complete flow (which is a part of both amqp and mqtt) should be good enough.

abhipsaMisra (approved):
lgtm

PR #54, opened 2021-02-04
Add runReason and exitReason to startingRun and endingRun events

BertKleewein:
I'm trying something new.  I'm firing events for different operations (like sending telemetry) and using a guid in the customDimensions to relate that to other events (like receiving telemetry).  The kusto join is fairly simple and gives us some better metrics.


vishnureddy17 (approved):
At this point, the PR is mainly reporting a run reason and an exit reason to Azure monitor rather than what you originally said. Otherwise, LGTM

PR #55, opened 2021-02-04
Don't log error and warning messages twice

BertKleewein:
Python logging, as configured, would push all warnings and errors _twice_ to Azure Monitor.  Info and Debug would only go once.  This filters out the duplicate messages.


vishnureddy17 (approved):
LGTM

PR #56, opened 2021-02-04
Fix duplicate serviceAckId bug

BertKleewein:
I had previously fixed this bug in one spot, but failed to find the root cause.  The fix in device.py only handled one kind of duplicate.  The cause was actually a bug in service.py.


vishnureddy17 (approved):
LGTM

PR #61, opened 2021-02-17
Initial Implementation of Node.js device app

vishnureddy17:
Currently implements C2D and telemetry testing.
Notably, it does not yet:
- Report system properties
- Report test configuration
- Report metrics and/or logs
- Test reported properties


nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+    clearInterval(this.telemetrySendingInterval);
+    clearTimeout(this.pairingAttemptTimer);
+    clearTimeout(this.pairingTimer);
+    if (this.twin) {
+      this.twin.removeAllListeners();
+    }
+    if (this.client) {
+      this.client.removeAllListeners();
+      this.client.close().catch(e => console.error(`Error closing client: ${e}`));
+    }
```
BertKleewein, line 267: If I'm reading this right, this calls close as an async operation and returns.  If close fails, it outputs to the console.  There is no way, from outside this function, to know _when_ the close operation succeeds.
Since it's calling an async function( `client.close()`), should `stopDeviceApp` also be an async function?

BertKleewein: I don't know if this is canon on the Node team, but I think every function, except for simple ones (getters/setters/transformers) should either accept a callback or return a promise.  (I assume that putting the `async` keyword in front of a function causes it to return a promise, right?)

---
In reply to: [577844534](https://github.com/Azure/iot-sdk-longhaul/pull/61#discussion_r577844534) [](ancestors = 577844534)

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+          cmd: 'serviceAckRequest',
+          serviceInstanceId: this.serviceInstanceId,
+          runId: runId,
+          serviceAckId: serviceAckId,
+          serviceAckType: 'telemetry'
+        }
+      }));
+      this.client.sendEvent(message)
+        .catch(() => {console.warn('Error sending message with serviceAckId: ' + serviceAckId)})
+        .finally(() => {this.serviceAckWaitList.add(serviceAckId)});
```
BertKleewein, line 246: Later on, we'll be counting these kinds of errors so we can report the count and fail the test if it gets too high

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) Microsoft. All rights reserved.
```
BertKleewein, line 1: global comment: I know that Tony is push using prettier to format code in the node repo.  We should do that here, following the same conventions.

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+              console.info('Pairing response timeout. Requesting again.');
+              this.twin.removeAllListeners('properties.desired.thief.pairing');
+              pairWithServiceAttemptRecursive().then(res).catch(e => rej(e));
+            }, 1000 * this.settings.pairingRequestSendIntervalInSeconds);
+          })
+          .catch(e => rej(new Error('Updating reported properties failed: ' + e.message)));
+      });
+    }
+
+    return new Promise<void>((res, rej) => {
```
BertKleewein, line 115: >(res, rej) [](start = 29, length = 10)

We should follow the same convention as the Node team, but I would prefer `resolve, reject` instead of `res, rej`.

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+      throw new Error('.register() did not return a promise when it should have.');
+    }
+    const registrationResult = await registrationResultPromise.catch(e => {throw new Error('DPS registration failed: ' + e.message)});
+    const connectionString = 'HostName=' + registrationResult.assignedHub + ';DeviceId=' + registrationResult.deviceId + ';SharedAccessKey=' + deviceKey;
+    const hubClient = Client.fromConnectionString(connectionString, IotHubTransport);
+    await hubClient.open().catch(e => {throw new Error('Opening client failed: ' + e.message)});
+    return hubClient;
+  }
+
+  private pairWithService() {
```
BertKleewein, line 55: >pairWithService [](start = 10, length = 15)

style question: if a function returns a Promise, should it be marked with the `async` keyword?  Because I think it should.  Otherwise I can't look at the signature and know what to do with the return value.  Or maybe it needs to be marked with the right typing to indicate that it returns a Promise.  That might be better.  Even though they're sorta the same thing, a linter would handle them differently.

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+      thief: {
+        pairing: {
+          requestedServicePool: requestedServicePool,
+          serviceInstanceId: null,
+          runId: runId
+        }
+      }
+    };
+
+    const pairWithServiceAttemptRecursive = () => {
```
BertKleewein, line 67: A function that returns a promise that is also recursive?  I kinda get this, but it hurts my head to try to reason about the behavior.  I understand what you're doing here, but I think you could just as easily do this iteratively, and I think it would be clearer.

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
BertKleewein, line 2: >Licensed [](start = 3, length = 8)

Global comment: Promise objects are good for converting old (callback-style) code into new (async/await) code.  Do they serve any other purpose?  It seems like anything you could write with Promise object could be more clearly written with async/await.  It also seems like your code is hodge-podge of Promises and async functions.  My intuition says you should be consistent (unless you have a reason to not be consistent), but I'm not sure if I'm missing something.

nodejs/device/src/index.ts
vishnureddy17: Right now, the `serviceAckId` gets added to to the `serviceAckWaitList` set regardless of whether the sending is successful.

So it ends up being caught when the set is too big because (presumably) the service never sends sends back a C2D message with that GUID. 

Is that not what we want?

nodejs/device/src/index.ts
vishnureddy17: Agreed

nodejs/device/src/index.ts
vishnureddy17: I didn't put a type annotation on it because TypeScript was already able to infer the return type as `Promise<void>`, but I'll add the return type to make it clear to the reader. 

Not sure if it's common to mark functions with the `async` keyword if it doesn't use `await`, I'll have to look into that.

nodejs/device/src/index.ts
vishnureddy17: Well the goal of `stopDeviceApp()` is that it clears everything from the event loop, causing Node to exit.
When it succeeds, you would know because Node exited without printing an error. Though it should print a success message anyway. I will do that. 

I think this also explains why returning a promise from the function isn't that useful.

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+    });
+    this.completeC2d(msg);
+  }
+
+  private rejectC2d(msg) {
+    this.client.reject(msg).catch(e => {console.warn('Error when rejecting C2D message: ' + e.message)});
+  }
+
+  private completeC2d(msg) {
+    this.client.complete(msg).catch(e => {console.warn('Error when completing C2D message: ' + e.message)});
```
BertKleewein, line 182: >catch(e => {console.warn('Error when completing C2D message: ' + e.message)} [](start = 30, length = 76)

This function has the side-effect of swallowing failures.  If `complete(msg)` throws an exception, it prints something to the log, but nothing else changes.  Maybe that's OK, but I can't say that I'm a big fan of ignoring failures like this.

vishnureddy17: I noticed that too, and I agree. Though we want the device app to continue running when errors like this happen.
What are your thoughts on incrementing a counter and stopping the device app if the counter gets too high? We could add a field to `thiefSettings` where the max value of the counter can be configured.

BertKleewein: that's exactly what I was thinking

---
In reply to: [577957666](https://github.com/Azure/iot-sdk-longhaul/pull/61#discussion_r577957666) [](ancestors = 577957666)

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+  private registerMessageListener() {
+    this.client.on('message', msg => {
+      let obj;
+      try {
+        obj = JSON.parse(msg.getData());
+      }
+      catch {
+        this.rejectC2d(msg);
+        return console.warn('Failed to parse received C2D message. Rejecting.');
+      }
```
BertKleewein, line 194: This is a really bad habit that we (as an industry) need to get out of.  Implied, but not obvious, is the fact that this lambda gets called inside the context of the SDK.  In Python, that would mean that this gets called from an SDK thread, but since Node is single threaded, the distinction is a little finer.

I'm going to write more about what I mean later.  Right now, I need to take care of some stuff.

nodejs/device/src/thiefSettings.d.ts
```diff
@@ -0,0 +1,14 @@
+export type thiefSettings = {
```
anthonyvercolano, line 1: >t [](start = 12, length = 1)

I would capitalize the first letter of this type (and interface) name.  Here and everywhere.

nodejs/device/src/index.ts
vishnureddy17: @YoDaMa @anthonyvercolano What are your thoughts on keeping this recursive vs redoing it iteratively? It's definitely possible to have a good iterative solution too, Bert came up with one.

nodejs/device/src/index.ts
vishnureddy17: @YoDaMa @anthonyvercolano I asked in Teams, but might as well ask in here as well. Is it a good idea to mark the function with the `async` keyword if there is no use of `await`? Or should adding a type annotation for the return value be sufficient?

nodejs/device/src/index.ts
vishnureddy17: After looking at it more, I think I'll do it iteratively

nodejs/device/src/index.ts
vishnureddy17: @BertKleewein In the python app we have `THIEF_ALLOWED_CLIENT_LIBRARY_EXCEPTION_COUNT` in the `device_run_config` dict. Should the equivalent configurable value in the Node app be used as a max value here?

nodejs/device/src/index.ts
```diff
@@ -0,0 +1,318 @@
...
+                console.info('Updating pairing reported props: %j', pairingAcceptPatch);
+                promisify(this.twin.properties.reported.update)(pairingAcceptPatch)
+                  .then(res)
+                  .catch(e => rej(new Error('Updating reported properties failed: ' + e.message)));
+              }
+            });
+
+            // If it's taking too long for a service app to respond to our pairing request,
+            // retry by recursively calling pairWithServiceAttemptRecursive()
+            this.pairingAttemptTimer = setTimeout(() => {
```
YoDaMa, line 105: there's a nice way to create a promisified setTimeout since all your code is already async await. Create a `wait` method that is just 

```js
function wait (time) {
    new Promise((resolve) => {
        setTimeout(resolve, time);
    })
}
```

YoDaMa: eh I take this back, for your use cases this is easier to comprehend.

PR #62, opened 2021-02-26
Implement logger, azure monitor logging and events

vishnureddy17:
I ended up writing a simple logger instead of using a third party logger.

Made some other small tweaks here and there.

### DIFF
--- a/nodejs/device/src/index.ts
+++ b/nodejs/device/src/index.ts
@@ -408,7 +408,7 @@ class DeviceApp {
         if (
           this.serviceAckWaitList.size > settings.sendMessageAllowedFailureCount
         ) {
-          return this.stopDeviceApp(
+          this.stopDeviceApp(
             new Error(
               `The number of service acks being waited on exceeds sendMessageAllowedFailureCount of ${settings.sendMessageAllowedFailureCount}.`
             )
--- a/nodejs/device/src/index.ts
+++ b/nodejs/device/src/index.ts
@@ -5,33 +5,78 @@ import {
   Signal,
   ThiefPairingProperties,
   ThiefC2dMessage,
+  AzureMonitorCustomProperties,
 } from "./types";
 import { TimeoutError } from "./timeout";
+import { Logger, LoggerSeverityLevel } from "./logger";
 import { Mqtt as ProvisioningTransport } from "azure-iot-provisioning-device-mqtt";
 import { Mqtt as IotHubTransport } from "azure-iot-device-mqtt";
 import { SymmetricKeySecurityClient } from "azure-iot-security-symmetric-key";
 import { ProvisioningDeviceClient } from "azure-iot-provisioning-device";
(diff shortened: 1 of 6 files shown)


nodejs/device/src/logger.ts
```diff
@@ -0,0 +1,94 @@
...
+      this.appInsightsEnabled = true;
+      this.appInsightsSeverityLevel = settings.appInsightsSeverityLevel;
+      this.appInsightsClient = settings.appInsightsClient;
+    }
+  }
+
+  info = this.log.bind(this, LoggerSeverityLevel.INFO);
+  warn = this.log.bind(this, LoggerSeverityLevel.WARN);
+  error = this.log.bind(this, LoggerSeverityLevel.ERROR);
+  critical = this.log.bind(this, LoggerSeverityLevel.CRITICAL);
```
BertKleewein, line 48: These belong in the constructor -- and that's where tsc puts them anyway.  (except in the constructor, you need to specify this.info, etc instead of letting it be implied)

vishnureddy17: In typescript the class members need to be declared. So you're saying that they should be declared without assignment and then assigned in the constructor?

BertKleewein: No, I was wrong.  I was translating to JS in my head and forgot about needing to declare members for type safety.

nodejs/device/src/index.ts
```diff
@@ -359,24 +395,34 @@ class DeviceApp {
...
+        logger.error(
+          `Error sending message with serviceAckId: ${serviceAckId}`
         );
       })
       .finally(() => {
         this.serviceAckWaitList.add(serviceAckId);
+        if (
+          this.serviceAckWaitList.size > settings.sendMessageAllowedFailureCount
+        ) {
+          return this.stopDeviceApp(
```
BertKleewein, line 411: Do you need the `return` here?  Isn't that deceptive because you're not really returning anything and the code should never reach the point of returning anyway.

BertKleewein, line 411: >return this.stopDeviceApp( [](start = 10, length = 26)

Also, as a philosophical point, I know why you're exiting the app here instead of throwing an exception, but what's the "_right_" way to return a failure from an interval?  In other words, what if we wanted this code to somehow "return" the exception (whatever that means) instead of just exiting the app?  What is the most node-ish way to do this?

BertKleewein (approved):
:shipit:

nodejs/device/src/index.ts
vishnureddy17: EventEmitter?

vishnureddy17: Promises could work. Something like this:
```
function failableInterval() {
    return new Promise((resolve, reject) => {
        const timer = setInterval(() => {
            // do stuff
            if (/*failed*/) {
                clearInterval(timer);
                reject(/*err*/);
            }
        }, timeout);
    }
}
```

PR #63, opened 2021-03-08
Add session metrics reporting

vishnureddy17:
I realized it's probably better to implement reported property testing before working on the other metrics, so I'll just get the session metrics in for now.

### DIFF
--- a/nodejs/device/src/duration.ts
+++ b/nodejs/device/src/duration.ts
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+export function msToDurationString(delta: number) {
+    // [D:]h:m:s[.UUUUUU]
+    delta = Math.trunc(delta);
+    let [ms, sec, min, hr] = [1000, 60, 60, 24].reduce((accumulator: number[], currentValue) => {
+        accumulator.push(delta % currentValue);
+        delta = (delta - delta % currentValue) / currentValue;
+        return accumulator;
+    }, []);
+    return (delta ? `${delta}:` : '') + hr + `:${min}:` + sec + (ms ? '.' + `${ms}`.padStart(3, '0') : '');
+};
--- a/nodejs/device/src/index.ts
+++ b/nodejs/device/src/index.ts
@@ -6,6 +6,7 @@ import {
   ThiefPairingProperties,
   ThiefC2dMessage,
   AzureMonitorCustomProperties,
+  SessionMetics,
 } from "./types";
 import { TimeoutError } from "./timeout";
 import { Logger, LoggerSeverityLevel } from "./logger";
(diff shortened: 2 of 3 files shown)


nodejs/device/src/duration.ts
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+export function msToDurationString(delta: number) {
+  // [D:]h:m:s[.UUUUUU]
+  delta = Math.trunc(delta);
+  let [ms, sec, min, hr] = [1000, 60, 60, 24].map((currentValue) => {
+    const mod = delta % currentValue;
+    delta = (delta - mod) / currentValue;
+    return mod;
+  });
```
BertKleewein, line 11: I appreciate the way you're trying to do this, and I think this kind of thinking can lead to some beautiful code, but this is a bit confusing:
1. `delta` and `mod` mean something mathematically, but mapping them to reality is difficult.  I think this is because the units change on every iteration.  `currentValue` is also confusing because it's not really the current value.  It's more like "units", but that's not quite right.  I'm not sure what I would call it.  "max field value" maybe, but that's just as bad.
2. using `map` isn't quite right either.  I think you're doing this a bit upside-down.  Instead of mapping `[10,60,60,24]` onto the parameter (`delta`), you're mapping `delta` onto `[10,60,60,24]`.  Or maybe I'm the one who's upside down.  Either way, I really needed to twist my neck to figure this out.  
3. `map` in the traditional map/reduce setting needs to be purely functional, but yours isn't.  By modifying `delta` inside the lambda, you confuse things.  You're assuming a sequential evaluation of the array, and you're assuming it evaluates from left-to-right.  Every execution of the lambda affects the next execution, and that forces the reader to mentally unroll the loop in their head.

Look at the `divmod` function.  I think it could really help here.

nodejs/device/src/index.ts
```diff
@@ -418,6 +444,11 @@ class DeviceApp {
   };
 
   private async startTestOperations() {
+    logger.info("Starting thief reported property sending.");
+    this.thiefPropertySendingInterval = setInterval(
```
BertKleewein, line 448: Not related to this change, but in the medium term, we still need to figure out if there is a better way to return failure from an interval function.

BertKleewein (changes requested):
:clock1:

nodejs/device/src/index.ts
vishnureddy17: Yep, I have a list of small design tweaks that I want to make and that's on it. 

For errors related to metrics and logging, would we even want to do anything other than logging the error and moving on? They aren't the things that we are trying to test, so I'm not sure we want them to cause the program to terminate.

vishnureddy17: Also, the fact that built-in JS promises don't have cancellation built in is kinda frustrating, because that would work wonderfully for this sort of thing.

nodejs/device/src/duration.ts
vishnureddy17: I tried looking for a `divmod` function in the standard libraries but didn't find one, unfortunately

BertKleewein (approved):
:shipit:

PR #64, opened 2021-03-09
set hub and serviceInstanceId to make app insights workbook work

BertKleewein:


nodejs/device/link.sh
```diff
@@ -0,0 +1,15 @@
+# !/bin/bash
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for
+# full license information.
+set -e 
+
+NODE_ROOT=$1
```
vishnureddy17, line 7: `NODE_ROOT` is the root of the Node SDK repo, right? I think something like `NODE_SDK_ROOT` would be more clear.

vishnureddy17 (approved):
LGTM besides that one comment on `link.sh`

PR #65, opened 2021-03-11
change asynchonous operations to better handle errors and add cancelation

vishnureddy17:
I'm using a new pattern for handling errors from and cancellation of asynchronous operations.

### DIFF
--- a/nodejs/device/src/index.ts
+++ b/nodejs/device/src/index.ts
@@ -121,8 +121,8 @@ class DeviceApp {
       );
     }
     const registrationResult = await registrationResultPromise.catch(
-      (e: Error) => {
-        throw new Error(`DPS registration failed: ${e}`);
+      (err: Error) => {
+        throw new Error(`DPS registration failed: ${err}`);
       }
     );
     logger.info(
@@ -136,18 +136,29 @@ class DeviceApp {
       IotHubTransport
     );
     logger.info("Connecting to IoT Hub.");
-    await hubClient.open().catch((e: Error) => {
-      throw new Error(`Opening client failed: ${e}`);
+    await hubClient.open().catch((err: Error) => {
+      throw new Error(`Opening client failed: ${err}`);
     });
     return hubClient;
   }
 
-  private pairingAttempt() {
-    let finished = false;
-    let cancel = () => {
-      finished = true;
-    };
+  private pairingAttempt(): [Promise<ThiefPairingProperties>, () => void] {
+    let cancel: () => void;
     const promise = new Promise<ThiefPairingProperties>((resolve, reject) => {
+      let cleanedUp = false;
+      const cleanUp = () => {
+        cleanedUp = true;
+        this.twin.removeListener("properties.desired.thief.pairing", listener);
+        clearTimeout(pairingAttemptTimer);
+      };
+      cancel = () => {
+        if (cleanedUp) {
+          return;
+        }
+        logger.warn("Pairing canceled.");
+        cleanUp();
+        reject(new Canceled());
+      };
       const listener = (received: ThiefPairingProperties) => {
         appInsights.defaultClient.trackEvent({
           name: "ReceivedPairingResponse",
@@ -165,12 +176,11 @@ class DeviceApp {
         logger.info(
           `Service app ${received.serviceInstanceId} claimed this device instance`
         );
-        this.twin.removeListener("properties.desired.thief.pairing", listener);
-        clearTimeout(pairingAttemptTimer);
+        cleanUp();
         resolve(received);
       };
       const on_timeout = () => {
-        this.twin.removeListener("properties.desired.thief.pairing", listener);
+        cleanUp();
         reject(
           new TimeoutError(
             "Pairing desired properties event not received in time."
@@ -182,34 +192,8 @@ class DeviceApp {
         on_timeout,
         1000 * settings.pairingRequestSendIntervalInSeconds
       );
-
-      const finish = () => {
-        logger.warn("Pairing canceled.");
-        this.twin.removeListener("properties.desired.thief.pairing", listener);
-        clearTimeout(pairingAttemptTimer);
-        reject(new Canceled());
-      };
-
-      cancel = () => {
-        if (finished) {
-          return;
-        }
-        finish();
-      };
-
-      if (finished) {
-        finish();
-      }
-    })
-      .then<ThiefPairingProperties>((received) => {
-        finished = true;
-        return received;
-      })
-      .catch<ThiefPairingProperties>((err) => {
-        finished = true;
-        return err;
-      });
-    return { promise, cancel };
+    });
+    return [promise, cancel];
   }
 
   private async pairWithService() {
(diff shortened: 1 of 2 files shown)


nodejs/device/src/index.ts
```diff
@@ -175,11 +178,38 @@ class DeviceApp {
...
+        clearTimeout(pairingAttemptTimer);
+        reject(new Canceled());
+      };
+
+      cancel = () => {
+        if (finished) {
+          return;
+        }
+        finish();
+      };
```
BertKleewein, line 198: You have 2 functions named `cancel` in here, one in the outer scope that sets finished=True and this one.  

If I'm reading this correctly, you return the first one to the caller of `pairingAttempt` and this one (and  `finish` immediately above) get created in local scope, aren't used, and then eventually get collected.  Is this right? 

Or, since this one doesn't have a `let` or `var` keyword, that implies that you're replacing the `cancel` variable in the outer scope.  But, if you've already returned the old value of `cancel`, replacing it doesn't benefit the user, does it? Or are there some execution rules that mean that the outer scope `cancel` already has the new value before `pairingAttempt` returns?

Closures and Promises sometimes hurt my brain, especially since each language has different rules about how they work.

nodejs/device/src/index.ts
```diff
@@ -175,11 +178,38 @@ class DeviceApp {
...
+    })
+      .then<ThiefPairingProperties>((received) => {
+        finished = true;
+        return received;
+      })
+      .catch<ThiefPairingProperties>((err) => {
+        finished = true;
+        return err;
+      });
+    return { promise, cancel };
```
BertKleewein, line 212: Confusing indent.  I think this needs to be one tabstop to the left of where it is.

nodejs/device/src/index.ts
```diff
@@ -175,11 +178,38 @@ class DeviceApp {
...
         on_timeout,
         1000 * settings.pairingRequestSendIntervalInSeconds
       );
-    });
+
+      const finish = () => {
+        logger.warn("Pairing canceled.");
+        this.twin.removeListener("properties.desired.thief.pairing", listener);
+        clearTimeout(pairingAttemptTimer);
+        reject(new Canceled());
```
BertKleewein, line 190: suggestion: you do these `removeListener` `clearTimeout` operations in a few places.  do you want to add something like:
```
const cleanup = () => {
        this.twin.removeListener("properties.desired.thief.pairing", listener);
        clearTimeout(pairingAttemptTimer);
}
```
and call that everywhere you clean up? 

The benefit here and now is not all that great, but it makes maintenance much much easier, especially if every "cancellable promise" has a `cleanup` function that always gets called before resolving or rejecting.

BertKleewein, line 190: >reject(new Canceled()); [](start = 8, length = 23)

It's confusing that `finished` the boolean and `finish` the function have two definitions for the word "finish".  In one case, it means "completed" and in the other case, it means "cancel".

nodejs/device/src/index.ts
vishnureddy17: This was an attempt to handle the case where cancel is called before the executor of the promise is run. However after some testing I think the executor gets called synchronously when the Promise constructor is called, so this might not actually be needed. 

Objects are references, we can replace the `cancel` property and it will be changed outside the function. 
Consider this example, which prints "hello".
```js
function test() {
    let obj = {func: () => {console.log("hey")}};
    process.nextTick(() => {obj.func = () => {console.log("hello")}});
    return obj;
}

let obj = test();
process.nextTick(() => {obj.func()});
```

nodejs/device/src/index.ts
vishnureddy17: I'm letting prettier call the shots here.

nodejs/device/src/index.ts
vishnureddy17: Oh yeah, I can just rename the `finish` variable to be `done`.

vishnureddy17: Good call.

BertKleewein:
Except you're not replacing a member variable.  You're doing this:
```
let test = () => {
    let func = () => {console.log("hey")};
    process.nextTick(() => {func = () => {console.log("hello")}});
    return { func }
};

let obj = test();
process.nextTick(() => {obj.func()});
```

vishnureddy17:
> Except you're not replacing a member variable. You're doing this: ``` let test = () => { let func = () => {console.log("hey")}; process.nextTick(() => {func = () => {console.log("hello")}}); return { func } }; let obj = test(); process.nextTick(() => {obj.func()}); ```

Oh yeah, you're right. And that one prints "hey" instead of "hello". For some reason, the cancellation behavior seemed to be working as I expected, though. I need to investigate.

nodejs/device/src/index.ts
vishnureddy17: Okay, so it's that the executor of the promise is called synchronously when the Promise constructor is called. This means that the assignment of `cancel` in the Promise executor happens before the `return`.

nodejs/device/src/index.ts
```diff
@@ -204,25 +218,30 @@ class DeviceApp {
     ) {
       appInsights.defaultClient.trackEvent({ name: "SendingPairingRequest" });
       logger.info("Updating pairing reported props: %j", pairingRequestPatch);
-      const patchReceived = this.onReceivePairingPatch();
+      let pairingAttemptPromise: Promise<ThiefPairingProperties>;
+      [
+        pairingAttemptPromise,
+        this.cancelPairingAttempt,
```
BertKleewein, line 224: I don't think this is going to hurt _too_ bad, but the way you save this function and never un-set it it means that the node collector will keep at least part of the `pairingAttempt` closure in memory forever.  This kind of leak is really insidious in Python -- you keep a function pointer, and the closure for that function contains an object, and that object links to other objects, and so the function pointer keeps a bunch of objects alive for much longer than they need to.  `this.cancelPairingAttempt` only has meaning while `pairingAttemptPromiose` is executing, so they seem pretty tightly bound to each other.  It feels strange to keep them in 2 different places.

nodejs/device/src/index.ts
```diff
@@ -248,123 +267,160 @@ class DeviceApp {
...
-  // For example, if we get index 5 and then 7, we assume 6 is lost.
-  private handleTestC2dMessage(obj: ThiefC2dMessage, msg: Message) {
-    if (
-      Number.isNaN(obj.thief.testC2dMessageIndex) ||
-      typeof obj.thief.testC2dMessageIndex !== "number"
-    ) {
-      this.client.reject(msg).catch(this.handleRejectC2dError);
-      return logger.warn("Issue with testC2dMessageIndex property. Rejecting.");
-    }
+  private handleC2dMessages(): [Promise<void>, () => void] {
```
BertKleewein, line 270: >handleC2dMessages [](start = 10, length = 17)

Here's where documentation might help.  I think this promise handles a single C2d message after it's been received.  The name implies that it handle more than one message.  I'm also a little unsure because the verb "handle" is very vague.  There's a school of thought that says "If you can't come up with a more specific verb, then that means that you can't describe what the code does, and maybe it needs to be refactored."  I don't think that's the case here, but "handle" as a verb is a smell that I watch out for.

nodejs/device/src/index.ts
```diff
@@ -379,85 +435,153 @@ class DeviceApp {
...
+                )
+              );
+            }
+          });
+      }, 1000 * settings.sendMessageOperationsPerSecond);
+    });
+    return [promise, cancel];
+  }
+
+  private sendThiefReportedProperties(): [Promise<void>, () => void] {
```
BertKleewein, line 506: >sendThiefReportedProperties [](start = 10, length = 27)

I keep hitting the same set of questions on all of these "top level" promises.  `sendThiefReportedProperties` implies a single send, but it looks like this is a recurring operation.  I don't know the answer, but what's the right way to differentiate between doing something once, and doing it over-and-over again?
3 ideas:
1. function-level documentaiton.
2. naming conventions.
3. is there a way to be more explicit using typings?  e.g. make a new type for this with a name like `RecurringCancellableTestOperation` and the function would be defined as ` private sendThiefReportedProperties(): RecurringCancellableTestOperation {`
(OK, that name is terrible and this might be _less_ readable now that I read it, but there might be an idea here)

nodejs/device/src/index.ts
```diff
@@ -379,85 +435,153 @@ class DeviceApp {
...
+    logger.info("Starting thief reported property sending.");
+    let cancel: () => void;
+    const promise = new Promise<void>((_, reject) => {
+      let cleanedUp = false;
+      let cleanUpAndReject = (err: Error) => {
+        cleanedUp = true;
+        clearInterval(interval);
+        reject(err);
+      };
+      cancel = () => {
```
BertKleewein, line 516: >cancel [](start = 6, length = 6)

exitReason is one of the reported properties.  The python code pushes reported properties one last time after cancel so the exit reason can be pushed up.  Depending on the failure, the push might not succeed, but it almost always does (in python at least).

BertKleewein (approved):
:shipit:

PR #66, opened 2021-03-16
Implement reported properties testing

vishnureddy17:
And address some feedback items from the prior review

### DIFF
--- a/nodejs/device/src/index.ts
+++ b/nodejs/device/src/index.ts
@@ -7,6 +7,7 @@ import {
   ThiefC2dMessage,
   AzureMonitorCustomProperties,
   SessionMetics,
+  TestReportedProperties,
 } from "./types";
 import { Canceled, TimeoutError } from "./errors";
 import { Logger, LoggerSeverityLevel } from "./logger";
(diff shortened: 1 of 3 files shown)


nodejs/device/src/index.ts
```diff
@@ -478,32 +478,53 @@ class DeviceApp {
...
         );
         logger.info(`Sending message with serviceAckId: ${serviceAckId}`);
-        this.client
-          .sendEvent(message)
-          .catch(() => {
-            logger.error(
-              `Error sending message with serviceAckId: ${serviceAckId}`
+        this.client.sendEvent(message).catch(() => {
+          logger.error(
+            `Error sending message with serviceAckId: ${serviceAckId}` //TODO: add a counter for these errors?
```
BertKleewein, line 483: >//TODO: add a counter for these errors? [](start = 71, length = 39)

also maybe remove from serviceAckWaitList?  I think so, but I might be convinced otherwise.

BertKleewein (approved):
:shipit:

nodejs/device/src/index.ts
vishnureddy17: So I was wondering what you're thoughts were on this. Pasting this message from Teams:

Suppose that the reported properties update function gives us an error when we're doing the operations for reported properties testing. 
 
What are your thoughts on just logging the error without incrementing a counter?
The reason I ask is because we can still add the GUID to the waitlist so the app would eventually fail if the waitlist gets too big due to the service never receiving the GUIDs in the first place.

BertKleewein: We still want a counter, but probably a single counter for all client errors.  We also want to be able to graph failures over time.

---
In reply to: [595461925](https://github.com/Azure/iot-sdk-longhaul/pull/66#discussion_r595461925) [](ancestors = 595461925)

PR #72, opened 2021-04-13
rename python objects for readability

BertKleewein:


python/device/running_operation_list.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(self):
+        self.lock = threading.Lock()
+        self.list = {}
+
+    def make_event_based_operation(self):
+        """
+        Make and return a running opreation object which fires an event when it is complete.
+        """
+        e = EventBasedOperation(self)
```
vishnureddy17, line 100: nit: Does `e` stand for `event` here? I'd rename this to something that describes it's purpose better. Same feedback applies elsewhere in the code.

vishnureddy17 (approved):

PR #73, opened 2021-04-14
Use YAML as one source of truth for constants

BertKleewein:


python/device/device.py
```diff
@@ -177,147 +169,143 @@ def _configure_azure_monitor_metrics(self):
...
-            "All virtual memory used by the process",
+            Metrics.PROCESS_BYTES_IN_ALL_HEAPS,
+            "Size of all heaps for the device app, essentially 'all available memory'",
             "bytes",
         )
         self.reporter.add_integer_measurement(
-            MetricNames.PROCESS_PRIVATE_BYTES,
-            "Amount of non-shared physical memory used by the process",
-            "bytes",
+            Metrics.PROCESS_PRIVATE_BYTES, "Amount of private data used by the process", "bytes",
```
vishnureddy17, line 187: What's the difference between the `PROCESS_PRIVATE_BYTES` and `PROCESS_WORKING_SET_PRIVATE` metrics?

python/device/device.py
```diff
@@ -177,147 +169,143 @@ def _configure_azure_monitor_metrics(self):
...
         )
 
         # --------------------
         # SendMesssage metrics
         # --------------------
         self.reporter.add_integer_measurement(
-            MetricNames.SEND_MESSAGE_COUNT_SENT,
-            "Count of messages sent and ack'd by the transport",
-            "message(s)",
+            Metrics.SEND_MESSAGE_COUNT_SENT, "Number of telemetry messages sent", "mesages",
```
vishnureddy17, line 208: typo: "messages", not "mesages". This appears in a few places.

vishnureddy17 (approved):

python/device/device.py
BertKleewein: I think there is no difference, but .net has 2 different measurements for these.  I haven't dug as deeply as I would like into this -- the 2 will probably be collapsed into 1 here at some point.

PR #75, opened 2021-04-29
overhaul waitlist to add timeout and promise-based operations

vishnureddy17:


BertKleewein:
>         msg: Message

This is out of scope and I should have noticed this sooner, but having 2 different parameters with "Message" in the name is confusing.  I _think_ they both come from the same place, but one is a message object from our SDK and the other is the payload from that message.  I'm not sure why these are split out.

---
Refers to: nodejs/device/src/index.ts:328 in ef68651. [](commit_id = ef68651416c74d29e7a684ab12abe22220d7a408, deletion_comment = False)

BertKleewein (approved):
:shipit:

PR #76, opened 2021-05-06
Convert operations to promise based, add reported properties testing, change metrics to update on an interval

vishnureddy17:
This is a big PR, but hopefully the way I broke up the commits makes it easier to review.


BertKleewein (approved):
:shipit:

PR #78, opened 2021-05-13
Add code to test methods.  Still needs much polish.

BertKleewein:


python/device/device.py
```diff
@@ -786,12 +807,21 @@ def dispatch_incoming_message_thread(self, worker_thread_info):
                 ):
                     # We only inspect messages that have `thief/runId` and `thief/serviceInstanceId` set to the expected values
                     cmd = thief[Fields.CMD]
-                    if cmd == Commands.SERVICE_ACK_RESPONSE:
+                    if cmd in [Commands.SERVICE_ACK_RESPONSE, Commands.METHOD_RESPONSE]:
```
vishnureddy17, line 810: I feel like the definition of service_ack is starting to become unclear now. For example, if the command is `Commads.METHOD_RESPONSE`, does this mean that we aren't considering the message we receive back as a service ack?

python/device/device.py
```diff
@@ -786,12 +807,21 @@ def dispatch_incoming_message_thread(self, worker_thread_info):
...
+                                "Received {} message with {}".format(
+                                    cmd, thief[Fields.SERVICE_ACK_ID]
+                                )
+                            )
+                        else:
+                            logger.info(
+                                "Received {} message with {}".format(
+                                    cmd, thief[Fields.SERVICE_ACKS]
+                                )
+                            )
```
vishnureddy17, line 824: Something like this might be cleaner
```suggestion
                        logger.info(
                            "Received {} message with {}".format(
                                cmd, thief[Fields.SERVICE_ACK_ID] if Fields.SERVICE_ACK_ID in thief else thief[Fields.SERVICE_ACKS]
                            )
                        )
```

python/device/device.py
```diff
@@ -921,6 +958,10 @@ def test_twin_properties_thread(self, worker_thread_info):
                 continue
 
             else:
+                self.subtest_method_invoke()
+                worker_thread_info.watchdog_epochtime = time.time()
+                time.sleep(self.config[Settings.TWIN_UPDATE_INTERVAL_IN_SECONDS])
+
```
vishnureddy17, line 964: Since there are are several different tests on this one thread, we might start running into issues there the tests start blocking each other too much.

You said that you are going to change how you do threading, so maybe you are already thinking about this.

python/device/device.py
```diff
@@ -1062,6 +1103,98 @@ def subtest_desired_properties(self):
...
+            method_request = self.client.receive_method_request(timeout=10)
+            if method_request:
+                logger.info("Received method request {}".format(method_request.name))
+                if method_request.name == MethodNames.FAIL_WITH_404:
+                    logger.info("sending response to fail with 404")
+                    self.client.send_method_response(
+                        MethodResponse.create_from_method_request(method_request, 404)
+                    )
+                elif method_request.name == MethodNames.ECHO_REQUEST:
+                    self.client.send_method_response(
```
vishnureddy17, line 1128: There's no log here, but you log when we send a 404 or 500 response. Not sure if you only wanted to log the non-success HTTP status codes or if this was just a mistake

python/service/service.py
```diff
@@ -188,6 +188,71 @@ def remove_device_from_pairing_list(self, device_id):
...
+                Fields.METHOD_INVOKE_CONNECT_TIMEOUT_IN_SECONDS, None
+            ),
+        )
+
+        # TODO: don't hold this lock while calling into registry manager
+        # TODO: wrap registry manager call in try/except to return result to device
+        logger.info("------------------------------------------------------------------")
+        logger.info("invoking {}".format(method_guid))
+        try:
+            with self.registry_manager_lock:
```
vishnureddy17, line 216: Why do we need a lock for the registry manager client?

vishnureddy17 (approved):

python/device/device.py
```diff
@@ -404,6 +424,7 @@ def update_initial_reported_properties(self):
         }
         self.client.patch_twin_reported_properties(props)
 
+    # TODO: rename props to payload or dikt
```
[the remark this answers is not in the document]
BertKleewein: "dict" is a reserved word in python.  many python libraries use "dikt" as a name when "dict" would be the more appropriate choice.

python/device/device.py
BertKleewein: I agree 100%.  I want to rename this as part of my refactor.  Maybe op_id?  I don't konw.

python/device/device.py
BertKleewein: Agree 100%.  that's part of my thinknig for the refactor.

python/device/device.py
BertKleewein: Good point.

python/service/service.py
BertKleewein: I don't like how I did this.  It's because I need to re-create the object when the token expires, and I destroy the old one in a fiarly nasty way (deleting internal objects and other things that make me feel dirty.).  I need to fix this.

PR #84, opened 2021-06-10
Move python send_message calls to non-critical threads

BertKleewein:


python/device/device.py
```diff
@@ -61,9 +61,9 @@
 # set default logging which will only go to the console
 
 logging.basicConfig(level=logging.WARNING)
-logging.getLogger("paho").setLevel(level=logging.DEBUG)
+# logging.getLogger("paho").setLevel(level=logging.DEBUG)
```
vishnureddy17, line 64: Is this meant to be commented out? Same with line 66

python/device/device.py
```diff
@@ -638,46 +559,47 @@ def test_send_message_thread(self):
         send_message_thread instance to actually send the message.
```
vishnureddy17, line 559: The description of `test_send_message_thread()` is outdated now

python/device/device.py
BertKleewein: Yes, I meant to leave these commented out

PR #85, opened 2021-06-10
refactor python twin/method loop

BertKleewein:


python/device/device.py
```diff
@@ -81,11 +81,7 @@
 TestMethod = collections.namedtuple(
     "TestMethod", "method_name expected_status_code include_payload"
 )
-methods_to_test = [
-    TestMethod(MethodNames.FAIL_WITH_404, 404, False),
-    TestMethod(MethodNames.ECHO_REQUEST, 200, True),
-    TestMethod(MethodNames.UNDEFINED_METHOD_NAME, 500, False),
-]
+methods_to_test = []
```
vishnureddy17, line 84: Can this list just be removed now?

python/device/device.py
```diff
@@ -772,26 +765,31 @@ def handle_incoming_test_c2d_messages_thread(self):
                         )
                     )
 
-    def test_twin_properties_thread(self):
+    def operation_test_thread(self):
         """
         Thread to test twin properties. The twin property tests consists of two subtests: one for
         reported properties and the other for desired properties.
         """
```
vishnureddy17, line 772: Seems like this description is outdated

python/device/device.py
```diff
@@ -583,19 +575,20 @@ def send_message():
...
+
+                with self.reporter_lock:
+                    self.reporter.set_metrics_from_dict(
+                        {
+                            Metrics.LATENCY_QUEUE_MESSAGE_TO_SEND: (send_time - queue_time) * 1000,
+                            Metrics.LATENCY_SEND_MESSAGE_TO_SERVICE_ACK: time.time() - send_time,
+                        }
+                    )
+                    self.reporter.record()
+            else:
```
vishnureddy17, line 590: I think incrementing a counter here and using that counter value in `check_failure_counts()` would be more consistent rather than using `send_message_count_not_received_by_service_app`

vishnureddy17 (approved):

python/device/device.py
BertKleewein: I think you're right.  I'm going to make this consistent with other ops in my next PR.

PR #92, opened 2021-06-20
Increase stress pressure and update metrics/settings

BertKleewein:
Stress updates:
* Increased default nubmer of operations per second and overlapped more operations.
* Split python thread pools into send & receive.  Under stress, there used to be a simple deadlock where all threads would be waiting for a response, so no responses could be handled.
* Added configuration options for token interval, keepalive, and thread counts.

Metrics/Settings:
* Limited memory metrics to working set and private working set.
* Added metrics for gc object count and active thread count.  These won't be used by all languages but are useful anyway.
* Removed latency metrics.  Because these measure round-trip and responses are batched, these are all but useless as-is.

### DIFF
--- a/python/device/device.py
+++ b/python/device/device.py
@@ -17,6 +17,7 @@ from executor import BetterThreadPoolExecutor, reset_watchdog, dump_active_stack
 import dps
 import queue
 import faulthandler
+import gc
 from azure.iot.device import Message, MethodResponse
 import azure.iot.device.constant
 from measurement import ThreadSafeCounter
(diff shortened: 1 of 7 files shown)


python/device/device.py
```diff
@@ -141,14 +142,18 @@ def __init__(self):
 Currently hardcoded. Later, this will come from desired properties.
 """
 device_run_config = {
-    Settings.MAX_RUN_DURATION_IN_SECONDS: 2 * 60,
+    Settings.MAX_RUN_DURATION_IN_SECONDS: 1 * 60 * 60,
```
vishnureddy17, line 145: I think you might have meant to put `1 * 60` here, not `1 * 60 * 60`

vishnureddy17 (approved):

PR #95, opened 2021-06-23
rename serviceAckId to operationId

BertKleewein:


python/device/device.py
```diff
@@ -102,7 +102,7 @@ def custom_props(extra_props):
 
 class CustomPropertyNames(object):
```
vishnureddy17, line 103: Why are these values here instead of in the constants file?

python/device/device.py
```diff
@@ -199,7 +199,7 @@ def __init__(self):
         self.metrics = DeviceRunMetrics()
         self.service_instance_id = None
         self.system_health_telemetry = SystemHealthTelemetry()
-        # for service_acks
+        # for running oeprations
```
vishnureddy17, line 202: *operations

vishnureddy17 (approved):

python/device/device.py
BertKleewein: I looked at how these are used, and we don't need these anymore.  I've removed them.

PR #99, opened 2021-07-02
Simplify pairing even more

BertKleewein:


python/device/device.py
```diff
@@ -511,64 +511,33 @@ def pair_with_service(self):
         while (time.time() - pairing_start_time) < self.config[
```
vishnureddy17, line 511: Looks like the description of `pair_with_service()` is outdated now.

vishnureddy17 (approved):

PR #101, opened 2021-07-16
add pytest scripts to test device client and track 2 POC SDK

BertKleewein:


python/sdk_e2e/test_track_1/conftest.py
```diff
@@ -0,0 +1,165 @@
...
+        except asyncio.TimeoutError:
+            pass
+        else:
+            logger.info("pairing response received")
+            msg = json.loads(running_op.result_message.data.decode())
+            return collections.namedtuple("ConnectedClient", "client service_instance_id")(
+                connected_client, msg[Fields.THIEF][Fields.SERVICE_INSTANCE_ID]
+            )
+
+    assert False
```
vishnureddy17, line 146: Maybe add a message here?

python/sdk_e2e/test_track_2/test_send_message.py
```diff
@@ -0,0 +1,83 @@
...
+
+        client.publish(
+            topic_builder.build_telemetry_publish_topic(message=test_message.message),
+            test_message.message.get_payload(),
+            qos=1,
+        )
+
+        assert not test_message.running_op.event.is_set()
+
+        # TODO: this test fails badly if we add this sleep here.  It's not just that it fails,
```
vishnureddy17, line 51: Why would this happen, and why is the sleep even useful for this test?

vishnureddy17 (approved):

python/sdk_e2e/test_track_2/test_send_message.py
BertKleewein: It's useful because it shows an order that I, as customer, could call the Paho APis in a way that can break it pretty badly.  If there is any time between calls to `pubish` and `reconnect`, it's broken in a way that it really hard to detect and hard to recover from.  (Unless you really understand the internals of Paho, but that kinda defeats the point ... )

This specific bug, with the sleep, is actually a bug in Paho.  But, if our strategy relies on customers callin Paho directly, it's something that we need to figure out.

PR #102, opened 2021-07-27
Add e2e tests for twin

BertKleewein:


python/service/service.py
```diff
@@ -519,23 +523,31 @@ def respond_to_test_content_properties(self, event):
...
 
                 with device_data.reported_property_list_lock:
                     if property_value:
                         operation_id = property_value[Fields.ADD_OPERATION_ID]
-                        device_data.reported_property_values[property_name] = property_value
+                        if Fields.REMOVE_OPERATION_ID in property_value:
+                            device_data.reported_property_values[property_name] = property_value
                     else:
-                        operation_id = device_data.reported_property_values[property_name][
+                        if (
```
vishnureddy17, line 535: Why can't this be an `elif`? Having an additional nesting here is less clean imo

vishnureddy17 (approved):
It might be worth trying to simplify logic in `service.py` to reduce the amount of nested blocks of code.

python/service/service.py
BertKleewein: I'll make a note to revisit that code.  I agree that it's confusing and I think it needs something more than de-nesting.  I'm just not sure what. :)

PR #103, opened 2021-07-30
add PNP tests

BertKleewein:


codegen/constants.yaml
```diff
@@ -159,6 +163,48 @@ groups:
           - name: METHOD_RESPONSE_STATUS_CODE
             value: "methodResponseStatusCode"
             desc: "Status code sent with method response"
+      - name: GetDigitalTwinC2d
```
vishnureddy17, line 166: Aren't we using "Plug and Play" instead of "Digital Twins"? 

IMO we should avoid using digital twins in our naming here because it can get confused with the Azure Digital Twins service.

python/common/thief_constants.py
```diff
@@ -22,6 +22,7 @@ class Fields(object):
     PROPERTIES = "properties"
     REPORTED = "reported"
     DESIRED = "desired"
+    RANDOM_CONTENT = "randomContent"
```
vishnureddy17, line 25: `thief_constants.py` doesn't seem to be in sync with `constants.yaml`

vishnureddy17 (approved):

PR #106, opened 2021-08-02
test methods and PNP commands

BertKleewein:


codegen/constants.yaml
```diff
@@ -233,7 +233,7 @@ groups:
           - name: UPDATE_PNP_PROPERTIES
             value: "updatePnpProperites"
             desc: "Update Pnp properites"
-          - name: INVOKE_PNP_CPOMMAND
+          - name: INVOKE_PNP_COMMAND
```
vishnureddy17, line 236: Sorry I didn't catch this earlier

python/sdk_e2e/test_track_1/conftest.py
```diff
@@ -244,7 +244,7 @@ def pnp_model_id():
 @pytest.fixture(scope="session")
 def random_key_factory():
     def factory_function():
-        return "prop{}".format(random.choice([1, 3, 4, 5]))
+        return "prop{}".format(random.choice([4]))
```
vishnureddy17, line 247: Isn't this equivalent to just doing `return "prop4"`?

python/service/service.py
```diff
@@ -212,7 +223,89 @@ def handle_method_invoke(self, device_data, event):
...
+        payload = thief.get(Fields.COMMAND_INVOKE_PAYLOAD, None)
+        response_timeout_in_seconds = thief.get(
+            Fields.COMMAND_INVOKE_RESPONSE_TIMEOUT_IN_SECONDS, None
+        )
+        connect_timeout_in_seconds = thief.get(
+            Fields.COMMAND_INVOKE_CONNECT_TIMEOUT_IN_SECONDS, None
+        )
+
+        logger.info(
+            "received command invoke command={}, componet={},  guid={}".format(
```
vishnureddy17, line 254: typo: `componet`

vishnureddy17 (approved):

python/sdk_e2e/test_track_1/conftest.py
BertKleewein: Yes, you are correct.  The property tests are very unreliable right now.  This is a temporary workaround that didn't actually work around anything, and I think I have a real fix in my my mental queue.

PR #109, opened 2021-08-04
move and re-organize fixtures

BertKleewein:


python/sdk_e2e/test_track_1/client_fixtures.py
```diff
@@ -0,0 +1,166 @@
...
+        except asyncio.TimeoutError:
+            pass
+        else:
+            logger.info("pairing response received")
+            msg = json.loads(running_op.result_message.data.decode())
+            return collections.namedtuple("ConnectedClient", "client service_instance_id")(
+                connected_client, msg[Fields.THIEF][Fields.SERVICE_INSTANCE_ID]
+            )
+
+    assert False
```
vishnureddy17, line 140: If messages are supported for assertions, I think a message here would be useful

vishnureddy17 (approved):

python/sdk_e2e/test_track_1/client_fixtures.py
BertKleewein: done.  changed to an exception with a message.

PR #110, opened 2021-08-05
create fixture for calling into service app

BertKleewein:


python/sdk_e2e/test_track_1/test_pnp_commands.py
```diff
@@ -4,8 +4,7 @@
 import pytest
 import logging
 import asyncio
-import json
-from thief_constants import Fields, Commands
+from thief_constants import Fields
```
vishnureddy17, line 7: Can't comment on the actual line, but is line 99 in this file supposed to be commented out?

vishnureddy17 (approved):

python/sdk_e2e/test_track_1/test_pnp_commands.py
BertKleewein: Yes  We can't actually get the status back from the twin service sdk.  I'm going to add a comment in my next PR and open a bug.  somewhere.  not sure where.

PR #112, opened 2021-08-09
rename running_operation to operation_ticket

BertKleewein:


python/sdk_e2e/test_track_1/test_twin.py
```diff
@@ -29,7 +29,7 @@ def validate_patch_reported(twin, patch):
     assert twin_prop == patch_prop
 
 
-# TODO: rename running_op to op_ticket?
+# TODO: rename op_ticket to op_ticket?
```
vishnureddy17, line 32: Remove this TODO

vishnureddy17 (approved):
I already said this on Teams, but `operation` is better than `op`, IMO. But that is minor, and this is WAY better than the previous name. 

I didn't catch anything like this, but there might have been something that was missed by find and replace so might want to watch out for that.

PR #113, opened 2021-08-09
utility to watch reported properties

BertKleewein:


scripts/watch_run.py
```diff
@@ -0,0 +1,34 @@
...
+import thief_secrets
+from thief_constants import Fields
+from azure.iot.hub import IoTHubRegistryManager
+
+registry_manager = IoTHubRegistryManager.from_connection_string(
+    thief_secrets.IOTHUB_CONNECTION_STRING
+)
+
+while True:
+    os.system("cls" if os.name == "nt" else "clear")
```
vishnureddy17, line 19: Not sure what the equivalent command is on Windows, but for Linux I think `clear -x` would be preferable so running the script doesn't clear the scrollback.

vishnureddy17 (approved):

vishnureddy17:
This is probably a lot of effort for minimal gain but it'd be cool if the changes between each `get_twin()` call were highlighted

scripts/watch_run.py
BertKleewein: that's good to know.   If this was more than a niggly little script that probably won't get used, I would probably use the blessings library because it does a much better job with full screen CLI

PR #114, opened 2021-08-09
fix pairing so test and longhaul both work again

BertKleewein:


python/service/service.py
```diff
@@ -331,6 +331,73 @@ def handle_pnp_command_invoke(self, device_data, event):
...
+
+            device_data = self.device_list.try_get(device_id)
+
+            # Handle the case where a device pairs with us
+            if not device_data and received_service_instance_id == service_instance_id:
+                device_data = PerDeviceData(device_id, received_run_id)
+                self.device_list.add(device_id, device_data)
+
+            # Handle the case where we used to be paired and we're not anymore
+            elif device_data and not received_service_instance_id == service_instance_id:
```
vishnureddy17, line 380: nit:
```suggestion
            elif device_data and received_service_instance_id != service_instance_id:
```

vishnureddy17 (approved):

python/service/service.py
BertKleewein: I agree. but the Python linter prefers the way it was in my PR and it's just not worth arguing with the linter....

PR #116, opened 2021-08-24
test python with reconnection disabled

BertKleewein:


python/sdk_e2e/test_track_1/test_sas_rewewal.py
```diff
@@ -0,0 +1,107 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+logger.setLevel(level=logging.INFO)
+
+pytestmark = pytest.mark.asyncio
+
+
+@pytest.mark.describe("Device Client with reconnect enabled")
+class TestSasRemewalReconnectEnabled(object):
```
vishnureddy17, line 15:
```suggestion
class TestSasRenewalReconnectEnabled(object):
```

python/sdk_e2e/test_track_1/test_sas_rewewal.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # and verify that we can send
+        await client.send_message(test_message.message)
+
+        # and verify that the message arrived at the service
+        await test_message.operation_ticket.event.wait()
+
+
+@pytest.mark.describe("Device Client with reconnect disabled")
+class TestSasRemewalReconnectDisabled(object):
```
vishnureddy17, line 63:
```suggestion
class TestSasRenewalReconnectDisabled(object):
```

python/sdk_e2e/test_track_1/test_sas_rewewal.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
vishnureddy17, line 1: Looks like these tests for reconnect enabled and reconnect disabled differ only in the `client_kwargs` fixture. Could you have a base class with the tests and then have two classes which inherit from it and override the `client_kwargs` fixture? It'd avoid repeating code.

vishnureddy17 (approved):

cartertinney (approved):

python/sdk_e2e/test_track_1/test_sas_rewewal.py
BertKleewein: Good idea.  I think I'm going to refactor to a base class with common tests in a future change.

